PR #21, opened 2022-12-05
Fix whitespace defect in status code definition

### LINKED ISSUE #19: Stray whitespace in the PSA status specification
According to the [PSA Firmware Framework 1.0.0](https://armkeil.blob.core.windows.net/developer/Files/pdf/PlatformSecurityArchitecture/Architect/DEN0063-PSA_Firmware_Framework-1.0.0-2.pdf), to the [master list predating PSA status as a separate specification](https://github.com/ARM-software/psa-ff-m/blob/master/allocated_errors.h), to [TF-M](https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/tree/interface/include/psa/error.h?h=TF-Mv1.6.1&id=1f6d1dd69193bdfe42c40bb3cec1d91b0cc73080) and to [Mbed TLS](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-2.28.0/include/psa/crypto_values.h#L44), the way to define a PSA status code is `#define PSA_ERROR_FOO ((psa_status_t)-42)`, with no spaces inside the definition.

According to the [PSA status code specification version 1.0](https://arm-software.github.io/psa-api/status-code/1.0/api/status-codes.html) and the [reference header `psa/error.h`](https://github.com/ARM-software/psa-api/blob/main/headers/status-code/1.0/psa/error.h), the way to define most PSA status codes is `#define PSA_ERROR_FOO ((psa_status_t) -42)`, with a space before the minus sign. There are a few exceptions, I think the ones that were imported from crypto in a later batch.

According to the C99 specification (n1256 §6.10.3 constraint 1, emphasis mine):

> Two replacement lists are identical if and only if the preprocessing tokens in both have the same number, ordering, spelling, and **white-space separation**, where all white-space separations are considered identical.

According to `gcc -Werror` or `clang -Werror`, the white-space separation bit does matter. Found when [testing PSA crypto](https://github.com/Mbed-TLS/mbedtls/pull/5998).

Conclusion: please make an erratum to the status code specification, removing the spaces before minus signs (and the `0` for `PSA_SUCCESS`). (I personally don't mind whether there's a space, but through historical happenstance, the PSA world has settled on no spaces.) And please note in the specification that this matters.

athoelke:
* Update API source
* Update API database
* Add change history

Fixes #19


gilles-peskine-arm (commented):
[Fix confirmed](https://github.com/Mbed-TLS/mbedtls/pull/5998).

The changes to the spec look good to me, but incomplete. Please add a note that implementations must provide compatible macros, down to the absence/presence of whitespace between tokens.

gilles-peskine-arm (approved):
LGTM

PR #27, opened 2022-12-12
Fix order of elements in Crypto API header

athoelke:
The order of the items in the reference header for the Crypto API was not "compiler-ready". Some of the type definitions are used in function declarations before they are defined.

The reference header is also missing the inclusion of psa/error.h.


ndevillard (approved):
Looks good

PR #36, opened 2022-12-16
Fix ambiguity in documentation of psa_xxx_setup() functions

### LINKED ISSUE #26: Possible ambiguity in documentation of `xxx_setup()` functions
The draft PAKE extension, in [the documentation of](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html#c.psa_pake_setup) `psa_pake_setup()`, states:

> If an error occurs at any step after a call to [psa_pake_setup()](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html#c.psa_pake_setup), the operation will need to be reset by a call to [psa_pake_abort()](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html#c.psa_pake_abort).

This seems to leave some ambiguity as to whether the user needs to call `psa_pake_abort()` when `psa_pake_setup()` returns an error, as there are two reasonable interpretation of that sentence:

1. The call to `psa_pake_setup()` is not a "step after a call to `psa_pake_setup()`", so is not covered by that statement, so the user might not be required to call `psa_pake_abort()` when `psa_pake_setup()` fails.
2. If `psa_pake_setup()` returns an error, that's an event that happens after `psa_pake_setup()` was called, so that statement applies to it and the user is required to call `psa_pake_abort()`.

Actually, all `xxx_setup()` functions seem to follow this pattern.

Note: @gilles-peskine-arm points out that [general conventions](https://armmbed.github.io/mbed-crypto/html/overview/conventions.html#behavior-on-error) so if `xxx_setup()` failed it leave the operation in an undefined state, so clearly if you want to re-use that operation object you need to call `xxx_abort()` on it, in order to bring it back to a defined state first - but what if you don't want to re-use it? Do you still need to call `xxx_abort()` on it (under possible pain of leaking resources or leaving your client-server setup in an inconsistent state)?

athoelke:
Clarify the behavior of the operation state after a call to an operation setup function. Add a state diagram to help illustrate the state model.

Fixes #26


athoelke:
You can see a preview of the diagram in the changed files list: doc/crypto/figure/multi_part_operation.svg

MarcusJGStreets (approved):

paul-elliott-arm (approved):
LGTM

PR #37, opened 2022-12-19
Clarify the hash parameter in hash signature functions

### LINKED ISSUE #17: Clarify the requirements for the hash parameter in calls to `psa_sign_hash()` and `psa_verify_hash()`
The specification of `psa_sign_hash()` and `psa_verify_hash()` does not place requirements to check that the input hash is valid for the signature algorithm.

Partly, this is due to the use in some protocols of separating the 'hash' step from the signing step; and where the input to the signing step is not the output of a cryptographic hash operation.

We propose to tighten the specification of these functions, to recommend that the implementation makes more effort to validate the input for algorithms where this is meaningful:

* Ensure that the format of the `hash` parameter is defined for all algorithms that can be used with these functions
* Add '`hash` is not a valid input value for the signature algorithm `alg`' to the list of conditions that can result in a `PSA_ERROR_INVALID_ARGUMENT` return code for these functions
* Explicitly require that `hash_length` must be equal to the output length of the associated hash function, for signature algorithms where the input to the signature **is** a hash.

athoelke:
Fixes #17


MarcusJGStreets (approved):

paul-elliott-arm (approved):
LGTM

PR #42, opened 2022-12-22
Remove URLs for CSTC standards

athoelke:
The location of documents published by CSTC changes from time to time. It is unhelpful to provide an obsolete URL with the citation reference.


ndevillard (approved):
Good idea. Thanks!

Issue #10, opened 2022-11-15
Firmware Update attributes: global or per-component?


athoelke:
In the 1.0-beta API, some aspects of the API behavior are global, for example [`PSA_FWU_MAX_WRITE_SIZE`](https://arm-software.github.io/psa-api/fwu/1.0/api/api.html#c.PSA_FWU_MAX_WRITE_SIZE), and some are per-component, for example [`PSA_FWU_FLAG_VOLATILE_STAGING`](https://arm-software.github.io/psa-api/fwu/1.0/api/api.html#c.PSA_FWU_FLAG_VOLATILE_STAGING).

The current spec implicitly permits an implementation to have different variations of the state model for different components (e.g. some might require a reboot to install, and others might not), although this is not explicitly stated.

Is there a reason for the attributes related to Firmware Update (such as these) to be global, instead of per-component, or per-component instead of global? - it seems we should have a good reason why some are global and others are not. There are some new attributes that might be in v1.0.0 - arising from #5, #8, and #9.

Currently, the per-component attributes are exposed via the component info structure, returned by `psa_fwu_query()` - and are runtime attributes. The global attribute is a build-time-constant macro. Macro definitions are clearly better for allocating buffers, and can reduce code footprint, but are less flexible for both implementations and clients.

athoelke:
Review of current FWU implementation- and/or component- specific attributes:

Attribute | Type | Comments
-|-|-
Maximum image buffer size | global macro |  `PSA_FWU_MAX_WRITE_SIZE`
Image block alignment | global macro | `PSA_FWU_WRITE_ALIGN`
Maximum image size | per-component value | `info.max_size`
Encrypted component image | per-component info flag | `info.flag & PSA_FWU_FLAG_ENCRYPTION`
Volatile staging | per-component info flag | `info.flag & PSA_FWU_FLAG_VOLATLE_STAGING`
TRIAL state used | implicit | Observe component state immediately after installation
Reboot required | implicit | Observe return value from `psa_fwu_install()`
Image verification | implicit | Is verification done during `psa_fwu_finish()`, `psa_fwu_install()`, and/or reboot? Consult implementation documentation, or code client defensively? - it could be any of these.
Manifest location | implicit | Is a component manifest detached or bundled with the image?

PR #48, opened 2023-01-20
Clarify the effect of volatile staging

### LINKED ISSUE #8: Clarify expected behavior when a component has transient staging
The initial implementation of the 1.0 API in TF-M v1.7 will not support persistent staging. Images being prepared for update (i.e. components in WRITING or CANDIDATE state) are not preserved when the system restarts. Support for this is planned in a later version of TF-M

This issue was identified as part of the TF-M implementation of the 1.0-beta API. The implementation patch and review is: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/17427.

Flexibility in the persistent/transient nature of the WRITING and CANDIDATE states is supported by the v1.0 API, but the specification provides no details on the expected impact on the state model, and implementation behavior.

### Proposal

Extend [Appendix C *Variation in system design parameters*](https://arm-software.github.io/psa-api/fwu/1.0/appendix/variations.html) to cover this use case.

Review other specification text to ensure that it is consistent with this use case.

### LINKED ISSUE #9: Support for bootloaders that do not report the outcome of an installation
### Context

The initial implementation of the 1.0 API in TF-M v1.7, using MCUboot is not able to report the result of an installation that occurs during reboot, when the outcome is not a TRIAL state.

This issue was identified as part of the TF-M implementation of the 1.0-beta API. The implementation patch and review is: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/17427.

The success or failure of the operation has to be deduced from the resulting version of the components after reboot:

* If the new images are active, then the installation was successful, and the old images have been discarded
* If the old images are active, then the installation failed, and the new images have been discarded

In both cases, this corresponds to READY state, and not to the UPDATED or FAILED states that are expected by the state model in the 1.0-beta API. See the state model variation for this use case in [Appendix C.1](https://arm-software.github.io/psa-api/fwu/1.0/appendix/variations.html#components-that-require-a-reboot-but-no-trial).

When TRIAL state is used, then a successful installation will result in TRIAL state, but a failure will result in READY state, not FAILED, as expected in the state model (see [Figure 7](https://arm-software.github.io/psa-api/fwu/1.0/overview/programming-model.html#fig-states)).

### Analysis

Options for the API:

1. Maintain the state model. This requires the Update service and bootloader to provide the missing information:

   - Enhance the bootloader to maintain extra states in the image metadata to model these states and report failure reasons
   - Store additional persistent metadata in the Update service to enable it to determine when a installation was attempted over a reboot, and what the outcome was

2. Relax the state model, and permit an implementation to optionally include the `clean` transition in an installation that occurs over reboot/restart.

   - This requires that on systems which skip the UPDATED/FAILED state in such transitions, the Update client (or Update server) evaluate the component version on device startup to determine if an attempted update has succeeded or failed.

TF-M would prefer option (2): option (1) incurs additional design and code complexity on their implementation to support additional persistent states, and they feel the burden on the Update client created by option (2) is acceptable.

Option (2) reduces the ability for the Update client at startup to determine its next action, as the API does not provide a report of installation success or failure. However, apart from not having an explicit reason for failure, the Update client (perhaps working with the Update server) is still able to determine the appropriate next action after the event.

This type of implementation can meet still meet the expectation that expensive erase operations do not occur in API calls except `psa_fwu_clean()`.

### Proposal

Relax the state model behavior for transitions over reboot, and permit a STAGED component to transition directly to READY on both success and failure of installation, as appropriate for the particular variant of the state model.

athoelke:
Add state diagrams to show the effect of volatile staging, and improve the description of the state model, and its variation.

Draft PDF preview: [IHI0093-PSA_Certified_Firmware_Update_API-1.0.0-pr48.pdf](https://github.com/ARM-software/psa-api/files/10969947/IHI0093-PSA_Certified_Firmware_Update_API-1.0.0-pr48.pdf)

Fixes #8
Fixes #9


Sherryzhang2 (commented):
I think this change is really helpful. Especially the change in no reboot and volatile case. After updating, no cleaning is required in this case and the component returns to the READY state directly.
Another comment is that what if one component is volatile while another is persistent in the system? It would be very complicated if dependency exists between them. How about adding descriptions like the components should be with the same attribute(persistent or volatile)?

athoelke:
@Sherryzhang2

> Another comment is that what if one component is volatile while another is persistent in the system?

This is partly the point of the question I asked in #10 - whether attributes should be per-component, or global. We have some of each right now, and the rationale for each choice is not necessarily obvious or clear.

>  It would be very complicated if dependency exists between them. How about adding descriptions like the components should be with the same attribute(persistent or volatile)?

This is worth thinking about. The effects of volatile staging are on the states involved in preparation and cleanup, rather than those involved in the installation operation itself. For example:

* If two components are both CANDIDATE when `psa_fwu_install()` is called and become STAGED, then a reboot will try to install both of them, irrespective of their volatility
* If the reboot happens before the call to `psa_fwu_install()`, then the volatile component would be discarded, but that does not present a dependency issue. After reboot, that component needs to be prepared again before installing both. A dependency would not affect the outcome.
* If the install fails, then the volatile component is reverted to READY (old firmware), and the persistent component will be FAILED. A dependency does not affect the result.
* If the install succeeds, then either both components move to TRIAL, or, if there is no trail, a volatile component is READY (new firmware) and the persistent one is UPDATED. A dependency does not affect the outcome.

In constrast, it is not possible to simultaneously (atomically) update two components, where one requires a TRIAL state and one does not support a TRIAL state.

athoelke:
I have added a preview of the document with this change to the PR description. The main changes are linked from the item in the Document change history appendix.

PR #49, opened 2023-01-27
Add alignment requirement for image writes

### LINKED ISSUE #5: PSA Firmware  API  FWU: support on flash nand with ECC
For platform having nand flash with ecc, the write access  need to be aligned on a minimum size which fluctuates according to platform.
Unaligned access on this plaform can be supported the FWU update service on device , but it adds complexity in this service and requires an additional block in flash to support unaligned access.  
Another solution is to  add an implementation define: PSA_FWU_WRITE_SIZE_UNIT 
On plaform without alignment constraint , the define is set to #define PSA_FWU_WRITE_SIZE_UNIT 1
for a plaform having 32 bytes aligment constraint, the value is set to #define PSA_FWU_WRITE_SIZE_UNIT 32
and the parameter   image_offset and  block_size from psa_fwu_write  are no more in bytes but in number of PSA_FWU_WRITE_SIZE_UNIT. 
 
NB: In Firmware_Update_API-1.0-bet.0, only one implementation define  (PSA_FWU_MAX_WRITE_SIZE)

athoelke:
Add PSA_FWU_WRITE_ALIGN to constrain the permitted image offsets, and permit implementation padding, when writing to a firmware image.

This enables efficient implementations on flash memories that have block-based write operations.

Fixes #5


MarcusJGStreets (approved):
Seems clear.
As far as I can tell, it does what we discussed.

bulislaw:
I'm a little sceptical about this change as it complicates the API further and blurs the abstraction layers.
Alignment requirement for an image is a static value and hardware specific detail. It is known to whoever is building firmware binary and plans update campaign upfront. The PSA FWU API shouldn't be used to enforce the alignment, at least in my opinion, as it spills more and more hardware specific details into the API layer and above.
I would move the necessary checks into the device specific implementation of the API, the part that actually knows where and how the HW stores the firmware, and if the update binary write violates the HW specific requirements then alignment error can be propagated up to the cloud.
Generally I think we should remember that the update package is build specifically for the device by a developer or integrator that knows all about the device and firmware.

athoelke:
> Generally I think we should remember that the update package is build specifically for the device by a developer or integrator that knows all about the device and firmware.

The issue here is not the way that that update image and package has been assembled, but the granularity in which the bytes of the update image are passed across the API. The conceptual model is that the Update client can be fully ignorant of the internals of an individual image (and they might be entirely opaque in the case of encrypted images).

If the client the update image data as a byte stream, and just transfers data as and when it is available to the Update service, this can result in a mismatch with the underlying storage blocks. This macro is to make the client to treat the image data as a sequence of blocks of size `PSA_FWU_WRITE_ALIGN`, as far as how the data is passed through the API.

athoelke:
If the implementation does partial block buffering, how does that interact with persistence of partially prepared images? - it means that every call to write must be persisted, even though every call might generate an unnecessary write (and then erase) cycle on some block. An implementation is mandated to provide an implementation of such persistence behavior (code and runtime costs), because the the API specification (without this alignment constraint) permits the client to pass the image one byte at a time.

Alternative approaches:
1. Permit the write function to fail if some [hidden] implementation requirement is not satisfied. I think this is worse than exposing the constraint in the API.
2. Require that any partial image write MUST be of size PSA_FWU_MAX_WRITE_SIZE, anything smaller is invalid, unless it is the final block of the image. (This satisfies the implementation problem of alignment, but forces the buffer size used by the client.
3. Have the implementation store the partial block in RAM, but know that the preparation of the image can not be restarted if there was any content in the RAM buffer. This either requires the existence of partial data to be persisted (i.e. probably worse solutoin than persisting the partial block), or for images in WRITING state to be discarded on boot because it is not known if some data was never persisted.
4. Do not permit implementations with persistent staging to have a block alignment issue in the firmware store.

bulislaw:
Let's take a step back. Writing to flash is usually constrained in some way. We are handling it currently in TF-M (or rather in mcuboot to be exact) in a way that's invisible to the API user. This involve reading existing data from Flash such that we can pad portion of the new where it doesn't neatly fit into program unit size. This ofc adds overhead in terms of time/power (up to two reads).

So I guess this request and change talks about allowing for optimisations.

My worries are as follows:
* We are jumping into changes to the API. I'm not saying we shouldn't handle it. I'm suggesting we need to do more research, understand the options better, prototype and get more partner's opinion; before committing to changing the API. This is not a must have feature and implications on the wider API should be considered carefully.
* Enforcing alignment means buffering data received from the network and constructing suitable blocks. No matter how and where we do it someone will need to keep track of what's downloaded and buffered and what's written if we are to handle reboots (as we should). We should research how download resumption is performed in cloud clients. Is it requested by the device (give me block whatever)? Does the cloud keep track of "acknowledged" packets and continues from there? There are different ways we could handle it from the lower layers, eg the fwu_write function could return number of bytes actually written to storage
* I'm not sure cloud clients can do anything better than buffer the data and pass it reanaged. The same can be done in the "hardware" layer where all other hardware quirks can also be considered and optimised for. 
* The alignment value is global for all the components, while in reality firmware for different components may be written to different storage backends (internal flash, external flash, external module even)
* Leaking hardware specific details to the generic upper layers feels bad. While we need to be pragmatic, this should be a  warning light to consider if the upper layer can do something better with the data and how it impacts the rest of the API.
* We are moving complexity to the adaptation layer that implements the PSA FWU API on the cloud client side, each cloud client integration will need to deal with it (or they may ignore it as "my platforms don't have this requirement"/"we dont care" which will lead to fragmentation).

jamike:
Hello, sorry for the response delay... The complexity is quite important and require additional persistent ressource : If block are not received align on PSA_FWU_MAX_WRITE_SIZE (size and offset) ,  an intermediate  persistent area (at least another block in flash must be reserved, possibly several according to the received data disordering and the number of simultaneous image update ) to write this data element  that can not be stored directly in secondary slot.  The Service firmware must implement the handling of the intermediate persistent area: at every received data, compute  aligned element that can be writen in secondary slot (assemble from intermediate persistent area) and store in intermediate persistent area the remaining part not matching alignment constraints.

athoelke:
This issue & PR was discussed during the working group call yesterday (20.4.23). A number of issues arose in the discussion:

1. This is not the only constraint that an implementation might reasonably want to impose on calls to `psa_fwu_write()`.

   * There can be good reasons for requiring the data to presented sequentially (e.g. for tracking progress),
   * or for wanting the input data buffer to be aligned (e.g. for DMA directly to a NVM controller)

2. Changes such as this one, or others that constrain the valid input to this function, could result in broken Update client code.
3. These constraints arise from the storage technology, different components' firmware can be stored in different locations, with different properties.

The current API definition implies that the Update client has total freedom in how it writes the data. In practice, this is not required, and is undesirable for almost any implementation.

One proposal is to specify a 'minimum' standard of behavior for implementations, that will satisfy current use cases, enabling portable Update client code. This could be based on the example in [Appendix B.3](https://arm-software.github.io/psa-api/fwu/1.0/appendix/examples.html#individual-component-update-multi-part-operation); or could also incorporate the alignment attribute to permit a client to supply smaller buffers that meet the alignment constraint.

Implementations are permitted to be more flexible in order to support more complex scenarios - and an Update client that depends on the non-standard behavior will be less portable across implementations. In future, with experience from such extensions, we might standardize these additional behaviors.

This approach would resolve the issues raised in (1), but permit implementation flexibility where possible and required. Although it would still induce a compatibility break _now_ (between 1.0-beta and 1.0.0), it would avoid compatibility issues that arise from handling other constraints.

To understand the effect on already written code (2), we should review the proposed changes (and alternatives) with developers of existing clients, and determine whether this change is an incompatibility, or if there are other approaches to the original issues that could work better.

The determination of the type of attribute (3) is related to the discussion in #10. A runtime attribute (e.g. provided in `psa_fwu_component_info_t`) has higher overheads to use, and does not permit compile-time optimisations, but might be necessary for attributes such as write alignment.

The idea of using a function-like macro to provide a compile-time mapping from component identifier to attribute value is not appealing: currently the mapping of component identifiers to firmware components for a specific implementation is not part of the API header file - some implementations of the API might leave that mapping to the system integrator.

athoelke:
Following a further discussion (at #5), the power-of-two constraint will be implemented by defining this value as the base-2 log of the alignment size. The limit of `17` will be a recommendation.

This also has the nice property that a value of zero literally indicates that no alignment is required.

athoelke:
Rebased, and added a new commit ([97ee309](https://github.com/ARM-software/psa-api/pull/49/commits/97ee309c1a3dd4c125ae074eb09a365933be69bf)), to use a Log2 value for the write alignment macro.

I've updated the [PDF in the description](https://github.com/ARM-software/psa-api/files/11545158/IHI0093-PSA_Certified_Firmware_Update_API-1.0.0-pr49v2.pdf), with the rendering of the full PR.

doc/fwu/api/api.rst
```diff
@@ -557,13 +557,40 @@ The following functions are used to prepare a new firmware image in the componen
...
+   *  Let :code:`WRITE_ALIGN_MASK = (1<<PSA_FWU_LOG2_WRITE_ALIGN) - 1`
+   *  If :code:`(image_offset & WRITE_ALIGN_MASK) != 0`, then the implementation returns :code:`PSA_ERROR_INVALID_ARGUMENT`.
+   *  If :code:`(block_size & WRITE_ALIGN_MASK) != 0`, then the implementation will pad the data with :scterm:`implementation defined` values up to the next aligned size, before writing the data to the firmware image.
+   *  This value does **not** constrain the alignment of the data buffer, ``block``.
+
+   The specific value of `PSA_FWU_LOG2_WRITE_ALIGN` is an :scterm:`implementation defined`, non-negative integer. If an implementation has no alignment requirement, then it defines `PSA_FWU_LOG2_WRITE_ALIGN` to be ``0``.
+
+   .. admonition:: Implementation note
+
+      It is recommended that `PSA_FWU_LOG2_WRITE_ALIGN` is not greater than ``17``, which corresponds to a block size of 128 KB.
```
bulislaw, line 577: Does it make sense to explain why?

bulislaw (approved):

doc/fwu/api/api.rst
athoelke: Should we explicitly mention the file delivery protocol that AWS uses for MQTT connections? - or can we have some more general wording about 'ensuring compatibility with existing block-based file transfer protocols used for IoT devices'?

PR #79, opened 2023-07-25
Clarifications prior to release

athoelke:
Some additional clarifications following final review:

* Improved the definition and references to 'volatile staging'
* Consistently use 'non-volatile' state/staging instead of 'persistent' state/staging
* Described the state volatility, and implications in the API definition, not just the programming model
* Consistently refer to a 'candidate image' rather than a mixture of 'candidate image', 'image in CANDIDATE state' and 'prepared image'
* Clarifications about the verification responsibilities of components

### DIFF
--- a/doc/fwu/about.rst
+++ b/doc/fwu/about.rst
@@ -19,19 +19,15 @@
    .. reference-table:: Documents referenced by this document
       :sorted:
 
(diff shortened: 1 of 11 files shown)


bulislaw (approved):
LGTM

PR #91, opened 2023-08-08
Add TLS-1.2 EC J-PAKE-to-PMS KDF algorithm

### LINKED ISSUE #11: New algorithm: Ad-hoc KDF for EC J-PAKE in TLS 1.2
Add API elements for the algorithm and supporting macros for the KDF used with EC J-PAKE in the TLS 1.2.

This has already been included in the development branch of Mbed-TLS (see https://github.com/Mbed-TLS/mbedtls/pull/6115), following review with the Crypto API authors.

Todo:
* Add and document `PSA_ALG_TLS12_ECJPAKE_TO_PMS` and `PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE` API elements to the Crypto API
* Add encoding for `PSA_ALG_TLS12_ECJPAKE_TO_PMS` to Appendix B
* [optionally] Provide a code snippet to demonstrate its usage

The relevant Mbed-TLS changes are as follows:

``` c
/* The TLS 1.2 ECJPAKE-to-PMS KDF. It takes the shared secret K (an EC point
 * in case of EC J-PAKE) and calculates SHA256(K.X) that the rest of TLS 1.2
 * will use to derive the session secret, as defined by step 2 of
 * https://datatracker.ietf.org/doc/html/draft-cragie-tls-ecjpake-01#section-8.7.
 * Uses PSA_ALG_SHA_256.
 * This function takes a single input:
 * #PSA_KEY_DERIVATION_INPUT_SECRET is the shared secret K from EC J-PAKE.
 * The only supported curve is secp256r1 (the 256-bit curve in
 * #PSA_ECC_FAMILY_SECP_R1), so the input must be exactly 65 bytes.
 * The output has to be read as a single chunk of 32 bytes, defined as
 * PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE.
 */
#define PSA_ALG_TLS12_ECJPAKE_TO_PMS            ((psa_algorithm_t)0x08000609)

/* The size of a serialized K.X coordinate to be used in
 * psa_tls12_ecjpake_to_pms_input. This function only accepts the P-256
 * curve. */
#define PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE 32
```

athoelke:
Fixes #11

Note that this does not exactly follow the MBEDTLS implementation:

*  The output size is provided by `PSA_TLS12_ECJPAKE_TO_PMS_OUTPUT_SIZE` rather than `PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE`, which is more consistent with the other Crypto API support macros.

### DIFF
--- a/doc/crypto/api.db/psa/crypto.h
+++ b/doc/crypto/api.db/psa/crypto.h
@@ -142,6 +142,7 @@ typedef /* implementation-defined type */ psa_mac_operation_t;
 #define PSA_ALG_SHA_512_256 ((psa_algorithm_t)0x0200000d)
 #define PSA_ALG_SM3 ((psa_algorithm_t)0x02000014)
 #define PSA_ALG_STREAM_CIPHER ((psa_algorithm_t)0x04800100)
+#define PSA_ALG_TLS12_ECJPAKE_TO_PMS ((psa_algorithm_t)0x08000609)
 #define PSA_ALG_TLS12_PRF(hash_alg) /* specification-defined value */
 #define PSA_ALG_TLS12_PSK_TO_MS(hash_alg) /* specification-defined value */
 #define PSA_ALG_TRUNCATED_MAC(mac_alg, mac_length) \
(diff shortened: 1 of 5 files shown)


athoelke:
@gilles-peskine-arm @mpg - Note also that in mbedtls:
1. The header documents `PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE` macro as the input size for the algorithm, but references it from the algorithm definition as the size of output.
2. The implementation uses `PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE` for both the output of the KDF, **but also for the input**.
 
It *just happens* to work because both are 32 bytes.

mpg:
@athoelke I'm going on holiday for 4 weeks at the end of the week, and have a number of urgent things that require my attention, so it's unlikely I'll find time to review before I leave, sorry. I've added this to my list, and will look at it when I return if I can't find time this week.

athoelke:
@mpg - would you be able to review this PR please?

mpg:
Thanks for the reminder, sorry I had forgotten about this, I'm having a look right now.

mpg (approved):
LGTM

PR #94, opened 2023-08-10
Add support for zigbee algorithms

### LINKED ISSUE #14: New algorithm: Zigbee's block-cipher based hash (AES-MMO)
Zigbee defines a cryptographic hash based on a block-cipher using the Matyas-Meyer-Oseas (MMO) construction. See [*Zigbee Specification r21*](https://zigbeealliance.org/wp-content/uploads/2019/12/docs-05-3474-21-0csg-zigbee-specification.pdf) §B.6.

The MMO construction is general, requiring:
* a block cipher that has a key length equal to the block size
* a padding operation on the input message to align with the block size
* an IV (or salt) of block-size length

Zigbee specifies all of these details for its hash function based on MMO:
* AES-128 is used as the block cipher
* The padding operation is similar to that used for MD and SHA hashes to prevent length-extension attacks, and permitting messages of less than 2<sup>32</sup> bits in length
* The IV is set to 0 (all bits zero)

The existing Crypto API for hash algorithms is not parameterized or salted. So supporting this use case as a Crypto API hash algorithm requires a new hash algorithm identifier for the Zigbee-specified hash function.

### LINKED ISSUE #15: New algorithm: Zigbee's encryption and authentication block-cipher mode (CCM*)
Zigbee defines a block-cipher mode, CCM*, that is derived from the CCM mode used for authenticated encryption with additional data (AEAD). See [*Zigbee Specification r21*](https://zigbeealliance.org/wp-content/uploads/2019/12/docs-05-3474-21-0csg-zigbee-specification.pdf) §B.1.1 and §A.

CCM is already supported by the Crypto API: see [`PSA_ALG_CCM`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/aead.html#c.PSA_ALG_CCM).

CCM* differs from CCM in that it additionally supports a construction of CCM which only provides encryption, i.e. authentication tag length *M* = 0, while retaining the CCM construction of the encryption stream from the nonce, configuration parameter *L*, and CTR mode.

Zigbee constrains the use of CCM* as follows:
* the block-cipher is AES-128
* the message-length encoding parameter, *L* = 2, limiting the plaintext length to < 2<sup>16</sup>
* the tag length *M* must be 0, 4, 8 or 16

Further, the construction of the 13 byte nonce is specified precisely, with the final byte encoding the value of *M* that is used (as part of the packet's 'security level'). See §4.5.2.2 in the Zigbee specification.

Zigbee's packet security can also optionally encrypt the payload. This is achieved by presenting the payload as part of the additional data for CCM* when encryption is not indicated by the security level, or as the plaintext for CCM* when encryption is required. See §4.4.1.1 and §4.4.1.2.

athoelke:
Add support for algorithms required by zigbee:
* MMO hash construction over AES-128
* CCM* optionally-authenticated variant of CCM

This PR adds a new hash function, and an unauthenticated cipher for the zero-length tag variant of CCM*. For non-zero-length tags, the existing support for CCM is adequate.

Fixes #14
Fixes #15

### DIFF
--- a/doc/crypto/api/ops/aead.rst
+++ b/doc/crypto/api/ops/aead.rst
@@ -70,6 +70,10 @@ AEAD algorithms
 
     The CCM block cipher mode is defined in :RFC-title:`3610`.
 
+    .. subsection:: Usage in zigbee
+
+        `PSA_ALG_CCM`, and its truncated variants, can be used to implement CCM* for non-zero tag lengths. CCM* is required by the :cite-title:`ZIGBEE`. For unauthenticated CCM*, the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm can be used.
+
     .. subsection:: Compatible key types
 
         | `PSA_KEY_TYPE_AES`
--- a/doc/crypto/api/ops/aead.rst
+++ b/doc/crypto/api/ops/aead.rst
@@ -72,7 +72,12 @@ AEAD algorithms
 
     .. subsection:: Usage in Zigbee
 
-        `PSA_ALG_CCM`, and its truncated variants, can be used to implement CCM* for non-zero tag lengths. CCM* is required by the :cite-title:`ZIGBEE`. For unauthenticated CCM*, the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm can be used.
+        The CCM* algorithm is required by :cite-title:`ZIGBEE`.
+
+        *   `PSA_ALG_CCM`, and its truncated variants, can be used to implement CCM* for non-zero tag lengths.
+        *   For unauthenticated CCM*, with a zero-length tag, use the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm.
+
+        See also :ref:`Usage in Zigbee <using-ccm-star-no-tag>` under `PSA_ALG_CCM_STAR_NO_TAG`.
 
     .. subsection:: Compatible key types
 
--- a/doc/crypto/api/ops/aead.rst
+++ b/doc/crypto/api/ops/aead.rst
@@ -70,7 +70,7 @@ AEAD algorithms
 
     The CCM block cipher mode is defined in :RFC-title:`3610`.
 
-    .. subsection:: Usage in zigbee
+    .. subsection:: Usage in Zigbee
 
         `PSA_ALG_CCM`, and its truncated variants, can be used to implement CCM* for non-zero tag lengths. CCM* is required by the :cite-title:`ZIGBEE`. For unauthenticated CCM*, the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm can be used.
 
--- a/doc/crypto/api/ops/ciphers.rst
+++ b/doc/crypto/api/ops/ciphers.rst
@@ -150,6 +150,39 @@ Cipher algorithms
         | `PSA_KEY_TYPE_CAMELLIA`
         | `PSA_KEY_TYPE_SM4`
 
+.. macro:: PSA_ALG_CCM_STAR_NO_TAG
+    :definition: ((psa_algorithm_t)0x04c01300)
+
+    .. summary::
+        The CCM* cipher mode without authentication.
+
+    This is CCM* as specified in :cite-title:`IEEE-CCM` §7, with a tag length of 0. For CCM* with a nonzero tag length, use the AEAD algorithm `PSA_ALG_CCM`.
+
+    The underlying block cipher is determined by the key type.
+
+    The IV generated or set in the cipher API is used as the nonce in the CCM* operation. An implementation must support the default IV length of 13. Support for setting a shorter IV is optional.
+
+    The maximum message length that can be encrypted is dependent on the length of the IV. See `PSA_ALG_CCM` for details of this relationship.
+
+    .. subsection:: Usage in zigbee
+
+        The zigbee message encryption algorithm is based on CCM*. This is detailed in :cite-title:`ZIGBEE` §B.1.1 and §A.
+
+        *   For unauthenticated messages — when *M* = 0 --- the `PSA_ALG_CCM_STAR_NO_TAG` algorithm is used with an AES-128 key in a multi-part cipher operation. The 13-byte IV must be constructed as specified in `[ZIGBEE]`, and provided to the operation using `psa_cipher_set_iv()`.
+
+            An implementation of zigbee cannot use the single-part cipher functions, as these generate a random IV, which is not valid for the zigbee protocol.
+
+        *   For authenticated messages — when *M* ∈ {4, 8, 16} --- the :code:`PSA_ALG_AEAD_WITH_SHORTENED_TAG(PSA_ALG_CCM, tag_length)` algorithm is used with an AES-128 key, where ``tag_length`` is the required value of *M*. The 13-byte nonce must be constructed as specified in `[ZIGBEE]`.
+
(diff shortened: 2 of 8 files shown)


MarcusJGStreets (approved):
Looks OK

doc/crypto/api/ops/aead.rst
```diff
@@ -70,6 +70,10 @@ AEAD algorithms
 
     The CCM block cipher mode is defined in :RFC-title:`3610`.
 
+    .. subsection:: Usage in zigbee
```
gilles-peskine-arm, line 73: Editorial: “Zigbee” should be capitalized. (It used to be “ZigBee”, but [that stopped years ago](https://en.wikipedia.org/wiki/Talk:Zigbee#Capitalization:_Zigbee,_not_ZigBee); permanently-uncapitalized “zigbee” seems to be the organization logo, not the name of the protocol suite).

doc/crypto/api/ops/ciphers.rst
```diff
@@ -126,6 +126,39 @@ Cipher algorithms
...
+
+    The maximum message length that can be encrypted is dependent on the length of the IV. See `PSA_ALG_CCM` for details of this relationship.
+
+    .. subsection:: Usage in zigbee
+
+        The zigbee message encryption algorithm is based on CCM*. This is detailed in :cite-title:`ZIGBEE` §B.1.1 and §A.
+
+        *   For unauthenticated messages — when *M* = 0 --- the `PSA_ALG_CCM_STAR_NO_TAG` algorithm is used with an AES-128 key in a multi-part cipher operation. The 13-byte IV must be constructed as specified in `[ZIGBEE]`, and provided to the operation using `psa_cipher_set_iv()`.
+
+            An implementation of zigbee cannot use the single-part cipher functions, as these generate a random IV, which is not valid for the zigbee protocol.
```
gilles-peskine-arm, line 149: Single-part _encryption_ is out due to the non-random IV, but what's wrong with single-part decryption?

doc/crypto/api/ops/aead.rst
```diff
@@ -70,6 +70,10 @@ AEAD algorithms
...
+
+        `PSA_ALG_CCM`, and its truncated variants, can be used to implement CCM* for non-zero tag lengths. CCM* is required by the :cite-title:`ZIGBEE`. For unauthenticated CCM*, the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm can be used.
```
gilles-peskine-arm, line 75: Please add a hyperlink to the discussion of Zigbee usage under `PSA_ALG_CCM_STAR_NO_TAG`, since that discussion covers both CCM and CCM\*-no-tag.

doc/crypto/api/ops/ciphers.rst
```diff
@@ -126,6 +126,39 @@ Cipher algorithms
...
+
+        *   For authenticated messages — when *M* ∈ {4, 8, 16} --- the :code:`PSA_ALG_AEAD_WITH_SHORTENED_TAG(PSA_ALG_CCM, tag_length)` algorithm is used with an AES-128 key, where ``tag_length`` is the required value of *M*. The 13-byte nonce must be constructed as specified in `[ZIGBEE]`.
```
gilles-peskine-arm, line 151: Please also list `PSA_ALG_CCM` as usable when the tag length is 16. It's logically redundant, but not didactically redundant.

doc/crypto/api/ops/hashes.rst
```diff
@@ -73,6 +73,20 @@ Hash algorithms
 
     RIPEMD-160 is defined in :cite-title:`RIPEMD`, and also in :cite-title:`ISO10118`.
 
+.. macro:: PSA_ALG_AES_MMO_ZIGBEE
+    :definition: ((psa_algorithm_t)0x02000006)
```
gilles-peskine-arm, line 77: Minor: it might conceivably make sense to define a SHA1 variant that forbids inputs which might be exploiting known collision weaknesses, due to SHA1's persisting use in protocols such as git. (Currently we aren't in a hurry to do that because that algorithm is not used in our typical domain space, and can't be implemented on constrained devices due to the high code size.) If we ever do that then it would be slightly more convenient to use hash 6 (since SHA-1 is 5). So MMO_ZIGBEE could be 7 instead.

doc/crypto/api/ops/hashes.rst
```diff
@@ -73,6 +73,20 @@ Hash algorithms
...
+
+    .. summary::
+        The *zigbee* 1.0 hash function based on a Matyas-Meyer-Oseas (MMO) construction of AES-128.
```
gilles-peskine-arm, line 80: Editorial: “construction _on_ AES-128”? _over_ AES-128?

doc/crypto/api/ops/hashes.rst
```diff
@@ -73,6 +73,20 @@ Hash algorithms
...
+
+    zigbee specifies a cryptographic hash function based on the MMO hash construction using the AES-128 block cipher. This is defined in :cite-title:`ZIGBEE` §B.6.
```
gilles-peskine-arm, line 82: “MMO construction using AES-128” is ambiguous since this is a theory-level definition which leaves some parameters open (padding, IV). The formulation here doesn't make it very clear that we mean the specific parametrization in the Zigbee specification. I suggest something like

> This is the cryptographic hash function based on the Merkle–Damgård construction over a Matyas–Meyer–Oseas one-way compression function and the AES-128 block cipher with the parametrization defined in `` :cite-title:`ZIGBEE` `` §B.6.

gilles-peskine-arm (changes requested):
Correctness: looks good to me. I've checked against the Zigbee specification and the Mbed TLS draft implementation of `CCM_STAR_NO_TAG`, but not against the IEEE CCM\* specification.

Writing: I left a couple of requests and suggestions.

doc/crypto/api/ops/aead.rst
athoelke: I've corrected all references except the document citation: the spec itself uses the uncapitalized form for its title and publisher.

doc/crypto/api/ops/ciphers.rst
athoelke: Although, technically, the single-part decryption function can be used for an implementation of Zigbee, in practice requires that the caller concatenate the 13-byte nonce and the cipher-text into a single memory buffer before calling the API.

But maybe there is no harm in making the explanatory text more precise.

doc/crypto/api/ops/ciphers.rst
athoelke: Done

doc/crypto/api/ops/hashes.rst
athoelke: Ok - we can recode this as `0x02000007`.

This was not already defined in Mbed TLS, unlike `PSA_ALG_CCM_START_NO_TAG`.

doc/crypto/api/ops/hashes.rst
athoelke: Decided on 'using' - same as the full description that follows.

doc/crypto/api/ops/hashes.rst
athoelke: Thank you - I have adopted the revised wording.

gilles-peskine-arm (approved):
LGTM

athoelke:
Rebased for merge

PR #101, opened 2023-09-15
Add new key agreement function

### LINKED ISSUE #85: Key agreement interface may present the shared key material out of the secured space.
`psa_raw_key_agreement`  and `psa_key_derivation_key_agreement` use the pointer to key to store the result of the key exchange. In the application where the Diffie-Hellman key exchange are used, the result of the DH (regardless it is raw or derived) can be used for another cryptographic service like the block cipher or MAC. In my opinion, it is more consistent to use the `psa_key_id_t * key` as a parameter to drain the result of the DH operation, it is also more consistent as well.

athoelke:
Add `psa_key_agreement()` as a standalone function that outputs to a new derivation key.

Fixes #85


doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,90 @@ Key agreement algorithms
...
+    .. param:: const uint8_t * peer_key
+        Public key of the peer. The peer key must be in the same format that `psa_import_key()` accepts for the public key type corresponding to the type of ``private_key``. That is, this function performs the equivalent of :code:`psa_import_key(..., peer_key, peer_key_length)`, with key attributes indicating the public key type corresponding to the type of ``private_key``. For example, for ECC keys, this means that peer_key is interpreted as a point on the curve that the private key is on. The standard formats for public keys are documented in the documentation of `psa_export_public_key()`.
+    .. param:: size_t peer_key_length
+        Size of ``peer_key`` in bytes.
+
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the new key.
+        This function uses the attributes as follows:
+
+        *   The key type must either be `PSA_KEY_TYPE_NONE` (``0``), or `PSA_KEY_TYPE_DERIVE`. :issue:`(Is this ok, or should we allow other unstructured (or just raw?) key types to be returned? Or should we mandate exactly one of those options, because the function always returns KEY_TYPE DERIVE?)`
```
gilles-peskine-arm, line 144: > It is reasonably safe to use the output as an HMAC key or as a PASSWORD (for input to PAKE or password-stretching algorithms). Should we universally permit such key types?

Permit but not require.

> Use of the output as a symmetric block-cipher or stream-cipher keys is strongly discouraged (for some ciphers this could lead to key discovery). Should we universally deny such key types?

I think this should be forbidden.

Note that `NONE` doesn't make sense: we never want to create a key object whose type is 0.

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,90 @@ Key agreement algorithms
...
+        *   The key size must either be ``0``, or the size of the key agreement output, in bits. :issue:`(Is this ok, or should we mandate leaving as default zero? Or should we permit the secret to be truncated into the output key?)`
```
gilles-peskine-arm, line 145: > Is it acceptable for the function to be required to return the entire shared secret as the key value? Should we require, or even permit, the caller to specify a key size?

Since the key size is known independently, the API should allow bits=0 to mean “whatever you calculate”, like with `psa_import_key`. The API should also allow the correct size value and reject nonzero incorrect values.

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,90 @@ Key agreement algorithms
...
+
+            The output size, in bits, of the key agreement is :code:`8 * PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE(type, bits)`, where ``type`` is the type of ``private_key`` and ``bits`` is the bit-size of either ``private_key`` or the ``peer_key``.
```
gilles-peskine-arm, line 147: > Truncating the output is unnecessary, and risks making the shared secret discoverable.

I actually wonder. For example, after a key exchange on secp521r1, you get a 521-bit number which is close to uniform over the range 0..2^521-1, but when forced into 8-bit bytes, you get a 66-byte string where the first byte's top bits are 0. So if you use this value, it may well make sense to just remove the _first_ byte.

You can truncate by exporting the shared secret and fiddling with it in the application space. But the whole point of this new function is that the shared secret remains in the secure world when the crypto implementation has isolation.

I'm wary of having customized truncation because this is basically creating a Spectre gadget.

tranvantruonggit: I basically think we shouldn't do anything with the shared secret before the shared secret is derived. So we don't have the problem of the first bit being zero. The truncation shall be after the key derivation.

athoelke: I've updated the PR to only support using the entire shared secret as the output key, and requiring the attributes to specify either zero or the correct number of bits (as per `psa_import_key()` and `psa_copy_key()`).

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,91 @@ Key agreement algorithms
...
+
+        *   ``private_key`` does not have the `PSA_KEY_USAGE_DERIVE` flag, or it does not permit the requested algorithm.
+        *   The implementation does not permit creating a key with the specified attributes due to some implementation-specific policy.
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The following conditions can result in this error:
+
+        *   ``alg`` is not a key agreement algorithm.
+        *   ``private_key`` is not compatible with ``alg``.
+        *   ``peer_key`` is not a valid public key corresponding to ``private_key``.
+        *   The output key attribuets in ``attributes`` are not valid :
```
athoelke, line 180: Typo:

```suggestion
        *   The output key attributes in ``attributes`` are not valid :
```

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,91 @@ Key agreement algorithms
...
+        The key agreement algorithm to compute: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_RAW_KEY_AGREEMENT(alg)` is true.
+    .. param:: psa_key_id_t private_key
+        Identifier of the private key to use.
+        It must permit the usage `PSA_KEY_USAGE_DERIVE`.
+    .. param:: const uint8_t * peer_key
+        Public key of the peer. The peer key must be in the same format that `psa_import_key()` accepts for the public key type corresponding to the type of ``private_key``. That is, this function performs the equivalent of :code:`psa_import_key(..., peer_key, peer_key_length)`, with key attributes indicating the public key type corresponding to the type of ``private_key``. For example, for ECC keys, this means that peer_key is interpreted as a point on the curve that the private key is on. The standard formats for public keys are documented in the documentation of `psa_export_public_key()`.
+    .. param:: size_t peer_key_length
+        Size of ``peer_key`` in bytes.
+
+    .. param:: const psa_key_attributes_t * attributes
```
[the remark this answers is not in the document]
gilles-peskine-arm: 1. That comes from our general principle that inputs go before outputs.
2. When there are multiple keys involved, we put the attributes close to the corresponding key, to avoid confusion.
3. I think `psa_key_derivation_output_key()` should have had the attributes next to the key. Putting them first doesn't make sense given that everywhere else the operation goes first. (That is because it would be the self parameter in an object-oriented language.)
4. The position makes sense to me. As with (2), there are no key parameters between the attributes and the corresponding key.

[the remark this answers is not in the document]
athoelke: Note that In (1) the key attributes comes before the other _input parameters_, but not next to the output key parameter.

Otherwise I agree with this analysis. If we could redo, I would put the attributes input parameter immediately before the key output parameter in all these cases (this is the only output for all these functions). I.e. I would change (1) and (3).

So keep the position in this PR, and use this approach in the other forthcoming key-creation functions.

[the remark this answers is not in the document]
gilles-peskine-arm: Possibly the order (1) for [`psa_import_key`](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/management.html#c.psa_import_key) is justified by the fact that the contents of the attributes affects how `data` is parsed, so it makes sense to put them before `data`.

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,91 @@ Key agreement algorithms
 Standalone key agreement
 ------------------------
 
+.. function:: psa_key_agreement
+
+    .. summary::
+        Perform a key agreement and return the shared secret as a derivation key.
+
+    .. param:: psa_algorithm_t alg
```
athoelke, line 130: Another parameter ordering consistency issue.

This API follows [`psa_raw_key_agreement()`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/ka.html#c.psa_raw_key_agreement), placing the algorithm before the key. Which happens to be the odd one out - _every_ other function that takes a key and an algorithm in this API, has the key parameter before the algorithm.

Should we make this API fit with the general pattern? - If so, is there any preference for _both_ keys coming before the algorithm ("Using these keys, compute key agreement `alg`"):

```
psa_status_t psa_key_agreement(psa_key_id_t private_key,
                               const uint8_t * peer_key,
                               size_t peer_key_length,
                               psa_algorithm_t alg,
                               const psa_key_attributes_t * attributes,
                               psa_key_id_t * key);
```

... or just the the `psa_key_id_t` parameter ("Using my key, compute key agreement `alg` with their key"):

```
psa_status_t psa_key_agreement(psa_key_id_t private_key,
                               psa_algorithm_t alg,
                               const uint8_t * peer_key,
                               size_t peer_key_length,
                               const psa_key_attributes_t * attributes,
                               psa_key_id_t * key);
```

... ?

gilles-peskine-arm: I suspect that the order of parameters of `psa_raw_key_agreement` comes from `psa_key_derivation_key_agreement`, which was written first. There the order is: operation, step, keys. The algorithm is extracted from the operation, so `psa_raw_key_agreement` has the algorithm first.

But I agree that it would have been better to put the algorithm in the same place as other functions that take an input key and an algorithm.

Either way, the input keys should be kept together. There's no reason to put the algorithm between `private_key` and `peer_key`.

athoelke: So that narrows it to two options: parameter consistency with `psa_raw_key_agreement()` or with 'all the other APIs'? - do I detect a preference for the latter? (That is my preference and leave raw key agreement as the odd one out)

MarcusJGStreets:
The latter, I would even think about deprecating psa_raw_key_agreement()
and replace it with a new function with the canonical argument order.


Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

doc/crypto/api/ops/ka.rst
gilles-peskine-arm: Yes, I prefer to leave `psa_raw_key_agreement` as the odd one out.

athoelke: I've reordered the parameters: see https://github.com/ARM-software/psa-api/pull/101/commits/18c73a6a8fefda442d5c20a28ae9752cd75827c0

doc/crypto/api/ops/ka.rst
```diff
@@ -146,6 +237,8 @@ Standalone key agreement
         The first ``(*output_length)`` bytes of ``output`` contain the raw shared secret.
     .. retval:: PSA_ERROR_INVALID_HANDLE
         ``private_key`` is not a valid key identifier.
+    .. retval:: PSA_ERROR_ALREADY_EXISTS
```
gilles-peskine-arm, line 240: This error is in the wrong list, it should be for `psa_key_agreement`.

doc/crypto/api/ops/ka.rst
```diff
@@ -168,11 +261,16 @@ Standalone key agreement
     .. retval:: PSA_ERROR_STORAGE_FAILURE
     .. retval:: PSA_ERROR_DATA_CORRUPT
     .. retval:: PSA_ERROR_DATA_INVALID
+    .. retval:: PSA_ERROR_INSUFFICIENT_STORAGE
```
gilles-peskine-arm, line 264: This error is in the wrong list, it should be for `psa_key_agreement`.

doc/crypto/api/ops/ka.rst
```diff
@@ -9,11 +9,19 @@
 Key agreement
 =============
```
gilles-peskine-arm, line 11: Missing change: in `doc/crypto/api/keys/policy.rst`: “This flag must be present on keys used with the following APIs: …”

gilles-peskine-arm, line 11: Missing changes: in `doc/crypto/appendix/encodings.rst`: “The former can only be used with …”

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,89 @@ Key agreement algorithms
...
+        Size of ``peer_key`` in bytes.
+    .. param:: psa_algorithm_t alg
+        The key agreement algorithm to compute: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_RAW_KEY_AGREEMENT(alg)` is true.
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the new key.
+        This function uses the attributes as follows:
+
+        *   The key type must be one of `PSA_KEY_TYPE_DERIVE`, `PSA_KEY_TYPE_RAW_DATA`, `PSA_KEY_TYPE_HMAC`, or `PSA_KEY_TYPE_PASSWORD`.
+
+            Implementations must support the `PSA_KEY_TYPE_DERIVE` and `PSA_KEY_TYPE_RAW_DATA` key types. Support for output as `PSA_KEY_TYPE_HMAC` or `PSA_KEY_TYPE_PASSWORD` is :scterm:`implementation defined` .
```
gilles-peskine-arm, line 145: This feels like an unnecessary allowance. I think we should just remove this sentence, and expect implementations to allow all the key types listed here unless they don't support a particular key type.

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,89 @@ Key agreement algorithms
...
+
+        *   The key size is always determined from the key agreement's shared secret. If the key size in ``attributes`` is nonzero, it must be equal to the size of the shared secret, in bits.
+
+            The output size, in bits, of the key agreement is :code:`8 * PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE(type, bits)`, where ``type`` is the type of ``private_key`` and ``bits`` is the bit-size of either ``private_key`` or the ``peer_key``.
```
gilles-peskine-arm, line 149: Preexisting: We shouldn't say “either `private_key` or `the `peer_key`”: that makes it look like the two keys can have different sizes and you can pick either. Instead, since `type` has to be the type of the private key, I think we should only mention `private_key` here. And in the entry for `peer_key`, we could state explicitly that it is parsed with the type `PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(type)` where `type` is the type of `private_key`, and with the same bit-size as the private key.

doc/crypto/api/ops/ka.rst
```diff
@@ -114,6 +122,89 @@ Key agreement algorithms
...
+        The key agreement algorithm to compute: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_RAW_KEY_AGREEMENT(alg)` is true.
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the new key.
+        This function uses the attributes as follows:
+
+        *   The key type must be one of `PSA_KEY_TYPE_DERIVE`, `PSA_KEY_TYPE_RAW_DATA`, `PSA_KEY_TYPE_HMAC`, or `PSA_KEY_TYPE_PASSWORD`.
+
+            Implementations must support the `PSA_KEY_TYPE_DERIVE` and `PSA_KEY_TYPE_RAW_DATA` key types. Support for output as `PSA_KEY_TYPE_HMAC` or `PSA_KEY_TYPE_PASSWORD` is :scterm:`implementation defined` .
+
+        *   The key size is always determined from the key agreement's shared secret. If the key size in ``attributes`` is nonzero, it must be equal to the size of the shared secret, in bits.
```
gilles-peskine-arm, line 147: Missing: saying explicitly that if the bit-size is specified as 0, it's the size of the shared secret.

doc/crypto/api/ops/ka.rst
gilles-peskine-arm, line 149: Wording: the text uses both “shared secret” and “output of the key agreement” indifferently. The two are synonymous. This makes the text hard to follow. Please pick one term except when explaining that the two are the same.

doc/crypto/api/ops/ka.rst
gilles-peskine-arm, line 11: We now use “raw key agreement” to mean three different things:

* An algorithm like `PSA_ALG_ECDH`, as opposed to `PSA_ALG_KEY_AGREEMENT(..., ...)`
* `psa_raw_key_agreement` or `psa_key_agreement`, as opposed to `psa_key_derivation_key_agreement`
* `psa_raw_key_agreement`, as opposed to `psa_key_agreement` and `psa_key_derivation_key_agreement`

I wonder if we should revise this. The only one we can change is the name of the function `psa_raw_key_agreement`. Unfortunately, the first two are the ones that make the more sense and that are closely related (a raw algorithm is what you pass to the “raw” functions, i.e. the functions that only do a key agreement and nothing else). Should we switch to a different adjective for “key agreement not combined with a key derivation”? Maybe “standalone”?

gilles-peskine-arm (changes requested):
The overall design looks good to me, but there are editorial mistakes and some places are unclear.

doc/crypto/api/ops/ka.rst
athoelke: Are you referring to just the last sentence, permitting impdef behavior; or the entire last paragraph, just leaving the statement about the permitted key types?

doc/crypto/api/ops/ka.rst
athoelke: I think these clarifications should also be applied (as relevant) to the other key agreement functions.

doc/crypto/api/ops/ka.rst
gilles-peskine-arm: Just the last sentence, i.e. having impdef for two of the key types.

doc/crypto/api/ops/ka.rst
athoelke: Happy to add that clarification.

Note that the wording here was derived from that used in `psa_import_key()` and `psa_copy_key()`:
* `psa_import_key()` says that the 'size is determined from the import data. If non-zero, ...' but does state in the full description the effect of a zero size in the attributes.
* `psa_copy_key()` says that the 'size can be zero. If non-zero, ...' with no further explanation anywhere.

An alternative would be to require that the size MUST be zero, as there is only one valid non-zero value, and it takes time and effort for the application to set it correctly, and for the implementation to validate it?

doc/crypto/api/ops/ka.rst
athoelke: Will do, and will review the rest of this section for the same terminology.

doc/crypto/api/ops/ka.rst
athoelke: Thanks. Evidence of use of cut and paste... 😀

doc/crypto/api/ops/ka.rst
gilles-peskine-arm: If the application knows exactly what the expected attributes are, it should be allowed to pass them. In particular, it should be possible to call `psa_get_key_attributes` and reuse the structure to create a new key with the same attributes (only changing the key id if one of the keys is persistent).

athoelke: Ok, fair use case.

doc/crypto/api/ops/ka.rst
athoelke: I have already used 'standalone key agreement' in the section title for `psa_key_agreement()` and `psa_raw_key_agreement()`. Elsewhere we already use 'combined key agreement' to describe the ka+kdf algorithms.

I think it makes sense to adjust terminology throughout the spec on this topic. There are a few other API instances, that we might want to reconsider:

* `PSA_ALG_IS_RAW_KEY_AGREEMENT()`
* `PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE()`
* `PSA_RAW_KEY_AGREEMENT_OUTPUT_MAX_SIZE`

Given that `psa_key_derivation_key_agreement()` has no 'output', we could add synonyms for these three macros, for example:

* `PSA_ALG_IS_STANDALONE_KEY_AGREEMENT()`
* `PSA_KEY_AGREEMENT_OUTPUT_SIZE()`
* `PSA_KEY_AGREEMENT_OUTPUT_MAX_SIZE`

athoelke: Thank you, good catch.

athoelke: Thank you. I had forgotten about these cross references.

athoelke: I've addressed the easy aspects in https://github.com/ARM-software/psa-api/pull/101/commits/ed15b615064c16339b3c53be22a6361eae1b2967, but would appreciate some discussion on whether we should also do something with the interface elements.

gilles-peskine-arm: Let's add `PSA_ALG_IS_STANDALONE_KEY_AGREEMENT()` as a synonym.

But we don't need to change buffer size macros since they aren't needed for `psa_key_agreement`, only to determine output buffer sizes for `psa_raw_key_agreement`.

athoelke: Added https://github.com/ARM-software/psa-api/pull/101/commits/b4869518646ddbe2af518286d2a160c8c4eb6212

gilles-peskine-arm (changes requested):
LGTM except for one problem with shared secret size vs key size.

doc/crypto/appendix/history.rst
```diff
@@ -69,7 +73,7 @@ Changes between *1.0.1* and *1.1.0*
 Changes to the API
 ~~~~~~~~~~~~~~~~~~
 
-*   Relaxation when a raw key agreement is used as a key's permitted-algorithm policy. This now also permits the key agreement to be combined with any key derivation algorithm. See `PSA_ALG_FFDH` and `PSA_ALG_ECDH`.
+*   Relaxation when a standalone key agreement is used as a key's permitted-algorithm policy. This now also permits the key agreement to be combined with any key derivation algorithm. See `PSA_ALG_FFDH` and `PSA_ALG_ECDH`.
```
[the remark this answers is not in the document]
athoelke: Thanks - fixed

doc/crypto/api/ops/ka.rst
```diff
@@ -108,42 +116,128 @@ Key agreement algorithms
...
+        The standalone key agreement algorithm to compute: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_STANDALONE_KEY_AGREEMENT(alg)` is true.
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the new key.
+        This function uses the attributes as follows:
+
+        *   The key type must be one of `PSA_KEY_TYPE_DERIVE`, `PSA_KEY_TYPE_RAW_DATA`, `PSA_KEY_TYPE_HMAC`, or `PSA_KEY_TYPE_PASSWORD`.
+
+            Implementations must support the `PSA_KEY_TYPE_DERIVE` and `PSA_KEY_TYPE_RAW_DATA` key types.
+
+        *   The key size is always determined from the key agreement's shared secret. If the key size in ``attributes`` is zero, the key is the size of the shared secret. If the key size in ``attributes`` is nonzero, it must be equal to the size of the shared secret, in bits.
```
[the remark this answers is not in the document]
athoelke: Please check that the rewording does not introduce other inaccuracies.

Issue #106, opened 2023-10-09
Add support for CMAC based KDF algorithms


adeaarm:
Current revision of the spec does not have support for CMAC based KDFs as described by NIST in [SP 800-108](https://nvlpubs.nist.gov/nistpubs/legacy/sp/nistspecialpublication800-108.pdf) in the list of KDF algorithms, but it support only HMAC based algorithms. This is important for use cases where CMAC used as the PRF in the KDF is fully accelerated by the underlying hardware platform. PRF can be used in different modes, ideally add support for all of them to extend flexibility for platform integrators.

MarcusJGStreets:
There are issues around CMAC, raised within SP 800-108 with mitigations. So we would need to ensure that the API enforced those. 

When using CMAC as the PRF in counter mode, a party with knowledge of the key KIN, who
also can freely select portions of the string Label || 0x00 || Context || [L] 2 (as described in
Appendix B), may be able to control the values of one or more of the K(i). One way to guard
against the possibility of this sort of key control would be by replacing the fixed input data
Label || 0x00 || Context || [L] 2
with
Label || 0x00 || Context || [L] 2 || K (0),
where
K(0) = PRF(KIN, Label || 0x00 || Context || [L] 2).


When using CMAC as the PRF in feedback mode, a party with knowledge of the key KIN, who
can also freely select portions of the string Label || 0x00 || Context || [L] 2 and/or portions of the
IV, may be able to control the values of one or more of the K(i), as was the case for counter
mode. One way to guard against the possibility of this sort of key control would be by requiring
that
IV = PRF (KIN, Label || 0x00 || Context || [L]2 )
and requiring the inclusion of a counter in the feedback mode (i.e., [i]2 is not optional).

When using CMAC as the PRF in double-pipeline mode, a party with knowledge of the key KIN
may be able to control the values of one or more of the K(i). One way to guard against the
possibility of this sort of key control would be by requiring the inclusion of a counter in double-
pipeline mode (i.e., [i] 2 is not optional).

athoelke:
SP 800-108 provides general construction methods for building KDFs, but does not define any specific KDF construction, that is adequate for implementing an interoperable KDF. We want to only define algorithm identifiers for precise algorithm specifications, that ensure compatibility of results across implementations.

HKDF and the HMAC-based TLS KDFs are defined because their behavior is fully specified.

[...]

PR #114, opened 2023-10-30
Change the PAKE output function to return a key

### LINKED ISSUE #86: It is impossible to derive multiple keys from the common secret in the PAKE API

The suggested PAKE interface passes the resulting raw common secret directly to a key derivation operation. This is good to avoid direct usage of the secret but it makes it impossible to derive more than one key from the secret because the key derivation operation can only be used once. Deriving multiple keys is mandatory for some protocols for instance to get a session key for the rest of the pairing phase and in addition a long term key to establish further sessions.

For example, HomeKit derives several temporary keys as well as the long term secret key from the SRP common secret. A single key derivation operation cannot be used because the separate keys are derived with separate salt and info values.

It would be better to return the raw secret as a key which can only be used as input to a key derivation operation. This would be easy to use and allows for multiple derived keys.

The same problem exists for the key agreement interface which is also directly coupled to a key derivation operation. Again there exists protocols where multiple keys are needed, for instance one for outgoing and one for incoming messages. In the case of key agreement, the `psa_raw_key_agreement()` function can often be used as a workaround but returning a derivation only key instead of the coupling to a key derivation operation would be a more general and cleaner solution. _[This related issue has also been raised as #85]_

### LINKED ISSUE #100: APIs for extracting the shared secret from PAKE operations
In the current beta PAKE API, we have `psa_pake_get_implicit_key()` which extracts the [unconfirmed] shared secret result into a single KDF operation object. With the introduction of SPAKE2+ (#73) which include confirmation of the secret in the protocol, and issues with the usability of the current API for some use cases (#86), it seems worth evaluating the design of this part of the API.

I also have concerns about the naming of this API, and by extension, the proposed `psa_pake_get_explicit_key()` in #73. To explain the current names, the qualifiers 'implicit' and 'explicit' are a indicator as to the authenticity of the key:

* The 'implicit key' is not authenticated. If the other party has used a different password, then the resulting shared secret is not common, and attempted communication using the 'implicit key' will fail - but detection of that failure depends on details of the following exchanges.
* An 'explicit key' is the result of a protocol that explicitly authenticates/confirms the shared secret from the exchange, before returning a value derived from the shared secret. Thus the protocol will not provide a shared value for use by the application unless it is confirmed that both parties have used the same password.

### 1. Flexible usage of the shared secret

Issue #86 suggests that we need an API that extracts the shared secret into a new key, similar in a way to how `psa_key_derivation_output_key()` can construct a key from a KDF. For full flexibility, this API could permit the partial use of the PAKE output, enabling the output to be used to construct a pair of keys. For example, a PAKE that outputs 256 bits could be split into two 128-bit AES keys by calling this output function twice.

If we provide such an API, how important is it to continue to provide a 'extract secret into key derivation operation' function as well? - this use case can be achieved by creating a volatile derivation key from the PAKE, inputting that key to a key derivation operation, and destroying the key [after the operation]. The case for providing the 'direct injection' API for JPAKE, was that the unconfirmed/implicit shared secret is not good for use as a pseudo-random key directly, due to bias in the output value - forcing the application to pass this through a KDF seemed like a good thing.

Would we ever need to permit an application to easily extract the shared secret data (i.e. without constructing an exportable key to do this)?

### 2. Separating unconfirmed outputs from confirmed outputs

I think it is unlikely that we would want a to expose the unconfirmed shared secret from a PAKE algorithm that included a confirmation phase in the protocol. For such protocols, the shared secret would only be available after confirmation. Are there any use cases that contradict such an approach?

If a protocol will either output an unconfirmed secret OR a confirmed secret, do we need two different functions (or function sets - see (1))? - the case to have different names is that it makes the situation clearer in the application: "does the application need to exchange some confirmation values prior to using the key for communication?" However, most application usage of PAKE algorithms is following higher level system specifications (e.g. Matter), which will define how to use the PAKE, key scheduling, and any confirmation protocol - so the hint provided by the API name has limited effect on mitigating the risk of a developer using an unconfirmed PAKE output.

So we could just have the singular `psa_pake_get_shared_key()` (or something similar), which will return whatever the protocol outputs. Developers need to check the algorithm/protocol documentation to determine if they need to do further confirmation.

If we want to retain separate APIs, so it is easier to see in the code what is being returned, I would like to propose that we use something like `psa_pake_get_unconfirmed_key()` and `psa_pake_get_confirmed_key()`, instead of the 'implicit' and 'explicit' qualifiers in the current API and proposal.

athoelke:
* Replace `psa_pake_get_implicit_key()` with `psa_pake_get_shared_key()`
* Add key confirmation attribute to the PAKE cipher suite

Fixes #100
Fixes #86


athoelke:

oberon-sk:
> * Replace `psa_pake_get_implicit_key()` with `psa_pake_get_shared_key()`
> * Add key confirmation attribute to the PAKE cipher suite

lgtm

> 1. Should the entire secret be contained in a single key output, or can it be split across multiple output keys?

As far as we understand, the goal of this PR is to export a shared secret and defer any key derivation of one or multiple keys into subsequent calls independent from the PAKE APIs. Then, the entire secret needs to be exported.

> 5. Which position should the key attributes parameter take in `psa_pake_get_shared_key()`? - there is no ideal answer as the precedents in the API are not consistent.

The `operation` should always be the first parameter, we regard `psa_key_derivation_output_key()` as an outlier.

athoelke:
> > * Replace `psa_pake_get_implicit_key()` with `psa_pake_get_shared_key()`
> > * Add key confirmation attribute to the PAKE cipher suite
> 
> lgtm

Thank you for your review

> > 1. Should the entire secret be contained in a single key output, or can it be split across multiple output keys?
> 
> As far as we understand, the goal of this PR is to export a shared secret and defer any key derivation of one or multiple keys into subsequent calls independent from the PAKE APIs. Then, the entire secret needs to be exported.

The concern is whether there are protocols, or PAKEs that expect the output to be split into two or more pieces as keys. For example, the way that SPAKE2+-draft02 splits the output of `Hash(TT)` into `Ka` and `Ke`. Note that is not a use case we need to support, and Matter's use of SPAKE2+ only uses the entire confirmed PAKE output as a key derivation key.

> > 5. Which position should the key attributes parameter take in `psa_pake_get_shared_key()`? - there is no ideal answer as the precedents in the API are not consistent.
> 
> The `operation` should always be the first parameter, we regard `psa_key_derivation_output_key()` as an outlier.

We have come to the same conclusion.

oberon-sk:
> The concern is whether there are protocols, or PAKEs that expect the output to be split into two or more pieces as keys. For example, the way that SPAKE2+-draft02 splits the output of Hash(TT) into Ka and Ke. Note that is not a use case we need to support, and Matter's use of SPAKE2+ only uses the entire confirmed PAKE output as a key derivation key.

We are not aware of a use case where this would be required either.

athoelke:
> 3. What key types should we permit or deny? - it seems that this is not appropriate for extracting keys that can require a variable amount of keying material (e.g. any DES, FFDH key pairs, Weierstrass ECC key pairs, RSA key pairs)

We would like to make this 'derivation keys only' (similar to the `psa_key_agreement()` PR #101), but need confidence that there are no use cases where the output is used directly for encryption, key wrapping (might need block-cipher-key), or AES-CMAC-based key derivation (would need an AES key).

> 4. Should the default key confirmation be: 'unconfirmed', 'confirmed', or 'whatever the default is for the selected algorithm'?

Currently leaning towards a preference of 'confirmed' by default (following a 'secure by default' principle), which means that algorithms like J-PAKE, without a confirmation phase, would require the application to explicitly request an unconfirmed key.

athoelke:
Added a missing update to the J-PAKE graphic, and reversed the definition of the key confirmation constants, so `PSA_PAKE_CONFIRMED_KEY` is now the default.

doc/ext-pake/api/pake.rst
```diff
@@ -343,9 +346,13 @@ A PAKE primitive is required when constructing a PAKE cipher-suite object, `psa_
 PAKE cipher suites
 ~~~~~~~~~~~~~~~~~~
 
-A PAKE algorithm uses a specific cryptographic primitive for key establishment, specified using a `PAKE primitive <pake-primitive>`. PAKE algorithms also require a cryptographic hash algorithm, which is agreed between the participants.
+Most PAKE algorithms have parameters that must be specified by the application. These parameters include:
+
+*   The cryptographic primitive used for for key establishment, specified using a `PAKE primitive <pake-primitive>`.
```
silabs-Kusumit, line 352: Typo

silabs-Kusumit:
> 1. Should the entire secret be contained in a single key output, or can it be split across multiple output keys?

If the key type of the output will be derivation only then multiple output keys are not needed.

> Added a missing update to the J-PAKE graphic, and reversed the definition of the key confirmation constants, so `PSA_PAKE_CONFIRMED_KEY` is now the default.

lgtm

> 2\. Related to (1): Do we need to provide something that evaluates the size of the shared key output?

If we are returning only a single output key then this may not be needed.

doc/ext-pake/api/pake.rst
athoelke: Sorry, I can't seem to see the problem. What am I missing?

silabs-Kusumit: > The cryptographic primitive used for for key establishment

Small thing

athoelke:
> 3. What key types should we permit or deny? - it seems that this is not appropriate for extracting keys that can require a variable amount of keying material (e.g. any DES, FFDH key pairs, Weierstrass ECC key pairs, RSA key pairs)

Even if we expect that the output should be for derivation only: we should permit implementations to output AES keys (as long as they use all of the shared secret?), as we expect some implementations will provide AES-CMAC based KDFs (see #106).

doc/ext-pake/api/pake.rst
athoelke: Thank you! The hardest to see error in things you have written yourself....

doc/ext-pake/api.db/psa/crypto-pake.h
```diff
@@ -27,22 +28,29 @@ typedef uint8_t psa_pake_step_t;
...
                           psa_algorithm_t hash_alg);
+void psa_pake_cs_set_key_confirmation(psa_pake_cipher_suite_t* cipher_suite,
+                                      uint32_t key_confirmation);
 void psa_pake_cs_set_primitive(psa_pake_cipher_suite_t* cipher_suite,
                                psa_pake_primitive_t primitive);
-psa_status_t psa_pake_get_implicit_key(psa_pake_operation_t *operation,
-                                       psa_key_derivation_operation_t *output);
+psa_status_t psa_pake_get_shared_key(psa_pake_operation_t *operation,
+                                     const psa_key_attributes_t * attributes,
+                                     psa_key_id_t * key);
```
yanesca, line 53: The situation here is very similar to the case of [key agreement](https://armmbed.github.io/mbed-crypto/html/api/ops/ka.html) schemes. There we have two APIs: `psa_raw_key_agreement()` and `psa_key_derivation_key_agreement()`. Outputing a key derivation object achieves the same as the latter. Outputing a key instead is somewhere between the two: the Crypto API implementation doesn't have full control over the use of the output, but doesn't release the output to the user either.

Outputting a key derivation object has the advantage that the implementation knows about the PAKE when inputting the key to the derivation object and can prevent misuse better or even adjust serialisation format if needed. If the PAKE output is released as a key, the only information the implementation can rely on is the length.

Why do we need to switch to outputting a key?

athoelke: > Why do we need to switch to outputting a key?

See #86. Use cases that use separate derivations to create multiple keys from the output of the PAKE. The existing API cannot be used in such systems, and maintaining the existing API alongside a new one to return a key does not appear to deliver substantial benefit for the application or implementation.

yanesca:
> Even if we expect that the output should be for derivation only: we should permit implementations to output AES keys (as long as they use all of the shared secret?), as we expect some implementations will provide AES-CMAC based KDFs (see https://github.com/ARM-software/psa-api/issues/106).

I think we should permit implementations to output any type of key that we might need for key derivation. If I understand the proposal in #106 correctly, this would mean `PSA_KEY_TYPE_DERIVE`, `PSA_KEY_TYPE_HMAC`, `PSA_KEY_TYPE_AES` and for the last two the key policy would be set to `PSA_ALG_SP800_108_COUNTER_HMAC ` and `PSA_ALG_SP800_108_COUNTER_CMAC` respectively.

yanesca:
> Should the entire secret be contained in a single key output, or can it be split across multiple output keys?

I think most of the PAKEs arrive at one shared secret and the output should be a single key. If a higher level protocol splits that before deriving keys, the partitioning of the key material should be part of the key derivation algorithm. For PAKEs that output several keys (like OPAQUE) we should either add a new API or require the application to call the key output function once more (and define which call produces which key), but we shouldn't output a concatenated key.

> Related to (1): Do we need to provide something that evaluates the size of the shared key output?

If we put the output in a key, the caller doesn't need to know the size upfront and can get it from the key attributes later, so I don't think we need to provide anything extra here.

athoelke:
> I think most of the PAKEs arrive at one shared secret and the output should be a single key. If a higher level protocol splits that before deriving keys, the partitioning of the key material should be part of the key derivation algorithm.

That is also my opinion, all other things considered.

> For PAKEs that output several keys (like OPAQUE) we should either add a new API or require the application to call the key output function once more (and define which call produces which key), but we shouldn't output a concatenated key.

If I understand you correctly, you suggest that for algorithms like OPAQUE, we might decide that the application must call `psa_pake_get_shared_key()` twice. However, the application does _not_ control how the output is split between those two calls: that is a product of the algorithm itself; in contrast to the idea in (1) where the application could control the splitting of the PAKE output into multiple key objects.

athoelke:
Rebased (following merge of #115), and updated the specification of key types that `psa_pake_get_shared_key()` can output (see https://github.com/ARM-software/psa-api/pull/114/commits/9d4f2c06cc737e029d19d3fb24e777daef8ea761).

Some PAKEs (e.g. J-PAKE) produce biased output, that is not suitable for use as an encryption key. Therefore:

* Only require that `PSA_KEY_TYPE_DERIVE` and `PSA_KEY_TYPE_HMAC` (see #123) are valid for all PAKE algorithms
* Permit additional key types (block-cipher keys) for PAKE algorithms that produce a pseudo-random shared secret.

doc/ext-pake/api/pake.rst
```diff
@@ -1084,8 +1088,14 @@ Multi-part PAKE operations
         The attributes for the new key.
         This function uses the attributes as follows:
 
-        *   The key type is required. It must be an unstructured key type that can be constructed directly from the PAKE shared secret. For example, :code:`PSA_KEY_TYPE_DERIVE`, :code:`PSA_KEY_TYPE_HMAC`, or :code:`PSA_KEY_TYPE_AES`.
-        *   The key size in ``attributes`` must be zero. The returned key size is always determined from the PAKE shared secret.
+        *   The key type is required.
+            All PAKE algorithms can output a key of type :code:`PSA_KEY_TYPE_DERIVE` or :code:`PSA_KEY_TYPE_HMAC`.
+            PAKE algorithms that produce a pseudo-random shared secret, can also output block-cipher key types, for example :code:`PSA_KEY_TYPE_AES`.
+
```
yanesca, line 1094: Is this empty line intentional?

yanesca (approved):
LGTM.

doc/ext-pake/api/pake.rst
athoelke: In this case, I think not, it was probably an artefact of cut and pasting the text below the blank line.

silabs-Kusumit:
LGTM

silabs-Kusumit (approved):

athoelke:
Rebased and merged.

PR #115, opened 2023-10-31
Combine `psa_pake_set_password_key()` with `psa_pake_setup()`

### LINKED ISSUE #89: The PAKE interface does not work easily with the Mbed TLS driver design

The suggested interface cannot be implemented in an opaque driver.

Opaque drivers are selected based on the key attributes provided. For a multi-part operation the driver is selected by the first function called. For PAKE this is `psa_pake_setup()`. However, no key is passed to this function. The only key involved is passed to the `psa_pake_set_password_key()` function which is called later and cannot be used for driver selection because the driver cannot be changed during a multi-part operation.

Unfortunately, the problem is not easy to solve. A change in the interface of the `psa_pake_setup()` function would work fine in most cases:

``` c
psa_status_t psa_pake_setup(
    psa_pake_operation_t *operation,
    const psa_pake_cipher_suite_t *cipher_suite,
    psa_key_id_t password,
    const uint8_t *user_id, size_t user_id_len,
    const uint8_t *peer_id, size_t peer_id_len,
    psa_pake_role_t role);
```

`role`, `user_id`, and `peer_id` are included because they are often needed to interpret the password value.

However, for some protocols the password hash cannot be calculated before some data is exchanged. For example, in a variant of SRP-6 the client first sends its public key to the server, the server then responds with the password salt and its own public key. The client therefore needs to calculate the public key before it receives the salt needed to calculate the password hash.

_This issue is worth considering as part of the addition of support for SPAKE2+ to the PAKE API. See #73_

athoelke:
Providing the password key at the point of setup enables an implementation to identify the correct driver/key location when setting up the operation, rather than deferring this decision until the application provides the password key. See #89 for discussion of the API design changes in this PR.

Fixes #89


athoelke:

silabs-Kusumit:
LGTM

yanesca:
LGTM

PR #118, opened 2023-11-03
KDF relaxations and clarifications

### LINKED ISSUE #112: Should a key derivation continue working after a verify failure?
In a nutshell: if [`psa_key_derivation_verify_bytes`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_verify_bytes) detects incorrect output (`INVALID_SIGNATURE` condition), may this abort the operation?

(Everything about `verify_bytes` here applies equally to `verify_key`.)

----

### Out-of-capacity state

Generally, when a [function fails in a multipart operation](https://arm-software.github.io/psa-api/crypto/1.1/overview/functionality.html#multi-part-operations), the multipart operation enters an error state.

The specification of [`psa_key_derivation_output_bytes`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_output_bytes) suggests, but does not clearly mandate a different behavior for `PSA_ERROR_INSUFFICIENT_DATA`:

> The operation’s capacity was less than output_length bytes. In this case, the following occurs:
>
> * No output is written to the output buffer.
> * The operation’s capacity is set to zero — subsequent calls to this function will not succeed, even with a smaller output buffer.

And in general:

> If this function returns an error status other than `PSA_ERROR_INSUFFICIENT_DATA`, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.

The intent is that on an `INSUFFICIENT_DATA` condition, the operation object should enter an out-of-capacity state. In that state, subsequent calls to functions that request more output (`output_bytes`, `output_key`, `verify_bytes`, etc.) will return `PSA_ERROR_INSUFFICIENT_DATA`, rather than `PSA_ERROR_BAD_STATE`. However the specification does not explicitly state that it would be wrong for an implementation to return `PSA_ERROR_BAD_STATE` for subsequent output calls in this scenario.

**Proposed action:** explicitly describe the out-of-capacity state and require that an `INSUFFICIENT_DATA` condition enters it (absent other errors). Applies to `psa_key_derivation_output_bytes`, `psa_key_derivation_output_key`, `psa_key_derivation_verify_bytes`, `psa_key_derivation_verify_key`.

Note that there is an edge case of requesting 0 bytes when the remaining capacity is 0. Should this succeeds because the whole requested output can be produced? Or should this fail so as not to make a distinction between capacity-0 and capacity-exceeded? In other words, if the initial capacity is 1 then what happens with the following sequence of calls?

1. `output_bytes(1)` → success
2. `output_bytes(0)` → success or `INSUFFICIENT_DATA`?
3. `output_bytes(0)` → success or `INSUFFICIENT_DATA`?

### Behavior after a failed verification

Generally, when a [function fails in a multipart operation](https://arm-software.github.io/psa-api/crypto/1.1/overview/functionality.html#multi-part-operations), the multipart operation enters an error state.

The specification of [`psa_key_derivation_verify_bytes`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_verify_bytes) suggests, but does not actually mandate different behavior on an `INVALID_SIGNATURE` condition:

> `PSA_ERROR_INVALID_SIGNATURE`
> 
> The output of the key derivation operation does not match the value in expected_output.
> 
> (…)
>
> If this function returns an error status other than `PSA_ERROR_INSUFFICIENT_DATA` or `PSA_ERROR_INVALID_SIGNATURE`, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.

Based on this sentence, and on the code snippet that approximates the intended behavior of `psa_key_derivation_verify_bytes`, it seems that the intent is that an `INVALID_SIGNATURE` condition consumes the output, but does not affect the state of the object.

However, the specification does not state this clearly. It seems to me that an implementation could make the operation object enter an error state when it reaches an `INVALID_SIGNATURE` condition.

Which behavior is desirable? I can see the appeal for an implementation to treat `verify_bytes` as a wrapper around `output_bytes` which does not otherwise affect the state of the object. However, this isn't quite right since a `PSA_ERROR_NOT_PERMITTED` condition does enter an error state, and the permission requirements are different for `output_bytes` and `verify_bytes`. Furthermore, it's inconsistent with the fact that the specification does require putting the operation in an error state if [`psa_key_derivation_verify_key`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_verify_key) fails to access the key to compare against (e.g. `PSA_ERROR_INVALID_HANDLE` condition).

From a usage perspective, I can't think of a use case where it's desirable to keep the operation going after an `INVALID_SIGNATURE` condition. It's somewhat uncommon to use the same inputs to both validate part of the output, and construct a key or other material with another part of the output. A typical case where it does happen is using the first half of the output of a password hashing function as a password hash, and the second half as a key encryption key). In this case, a verification failure indicates a wrong password, so the user is not authenticated and the putative key encryption key should not be used.

**Proposed action**: don't give any special treatment to `PSA_ERROR_INVALID_SIGNATURE`. Thus it will cause the operation to enter the error state. Applies to `psa_key_derivation_verify_bytes`, `psa_key_derivation_verify_key`.

It's not clear to me whether the specification intended to _mandate_ or _allow_ implementations to preserve the operation's validity on `INVALID_SIGNATURE`. So I'm not sure if this proposed action is merely restricting the permitted behaviors, or if it would be an incompatible change.

### LINKED ISSUE #113: Should psa_key_derivation_verify_bytes be allowed with a USAGE_DERIVE secret?
[`psa_key_derivation_input_key`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_input_key) on a secret input (`PSA_KEY_DERIVATION_INPUT_SECRET` or `PSA_KEY_DERIVATION_INPUT_PASSWORD`) that is passed as a key allows the following operations:

* `output_xxx` ([`psa_key_derivation_output_bytes()`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_output_bytes), [`psa_key_derivation_output_key()`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_output_key)) only if the secret key has the policy [`PSA_KEY_USAGE_DERIVE`](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/policy.html#c.PSA_KEY_USAGE_DERIVE).
* `verify_xxx` ([`psa_key_derivation_verify_bytes()`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_verify_bytes), [`psa_key_derivation_verify_key()`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_verify_key)) only if the secret key has the policy [PSA_KEY_USAGE_VERIFY_DERIVATION](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/policy.html#c.PSA_KEY_USAGE_VERIFY_DERIVATION).

You can't use a verify function on a key with only the DERIVE usage. This is similar to signature keys not automatically allowing verification. However, I'm not sure that's desirable. For backward compatibility, you might have a key created against the 1.0 API (which only had `USAGE_DERIVE`) and want to use it with the 1.1 API's `verify_xxx` functions.

Allowing DERIVE keys to be used as a secret for `verify_bytes` wouldn't change the security of the key, since you can simulate it with `output_bytes` followed by `memcmp`. Allowing DERIVE keys to be used as a secret for `verify_key` changes what you can effectively do with the keys involved, but I can't think of a scenario where this change would be problematic.

Are there good hygiene reasons to forbid `verify_bytes` and `verify_key` with a secret input that only allows `USAGE_DERIVE`?

athoelke:
*   Changed the policy for `psa_key_derivation_verify_bytes()` and `psa_key_derivation_verify_key()`, so that these functions are also permitted when an input key has the `PSA_KEY_USAGE_DERIVE` usage flag.
*   Removed the special treatment of `PSA_ERROR_INVALID_SIGNATURE` for key derivation operations. A verification failure in `psa_key_derivation_verify_bytes()` and `psa_key_derivation_verify_key()` now puts the operation into an error state.
*   Clarified the behavior of a key derivation operation when there is insufficient capacity for a call to `psa_key_derivation_output_bytes()`, `psa_key_derivation_output_key()`, `psa_key_derivation_verify_bytes()`, or `psa_key_derivation_verify_key()`.

Fixes #112
Fixes #113

### DIFF
--- a/doc/crypto/api/ops/kdf.rst
+++ b/doc/crypto/api/ops/kdf.rst
@@ -967,7 +967,7 @@ Key derivation functions
 
     However, calling `psa_key_derivation_verify_bytes()` works even if the key's policy does not permit output of the bytes.
 
-    If this function returns an error status other than :code:`PSA_ERROR_INSUFFICIENT_DATA` or :code:`PSA_ERROR_INVALID_SIGNATURE`, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.
+    If this function returns an error status other than :code:`PSA_ERROR_INSUFFICIENT_DATA`, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.
 
     .. note::
         Implementations must make the best effort to ensure that the comparison between the actual key derivation output and the expected output is performed in constant time.
@@ -1021,7 +1021,7 @@ Key derivation functions
 
     This is functionally equivalent to exporting the ``expected`` key and calling `psa_key_derivation_verify_bytes()` on the result, except that it works when the key cannot be exported.
 
-    If this function returns an error status other than :code:`PSA_ERROR_INSUFFICIENT_DATA` or :code:`PSA_ERROR_INVALID_SIGNATURE`, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.
+    If this function returns an error status other than :code:`PSA_ERROR_INSUFFICIENT_DATA`, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.
 
     .. note::
         Implementations must make the best effort to ensure that the comparison between the actual key derivation output and the expected output is performed in constant time.
--- a/doc/crypto/api/ops/kdf.rst
+++ b/doc/crypto/api/ops/kdf.rst
@@ -685,7 +685,7 @@ Key derivation functions
 
         *   `psa_key_derivation_output_bytes()` --- if each input was either a direct input or a key with usage flag `PSA_KEY_USAGE_DERIVE`.
         *   `psa_key_derivation_output_key()` --- if the input for step `PSA_KEY_DERIVATION_INPUT_SECRET` or `PSA_KEY_DERIVATION_INPUT_PASSWORD` was a key with usage flag `PSA_KEY_USAGE_DERIVE`, and every other input was either a direct input or a key with usage flag `PSA_KEY_USAGE_DERIVE`.
-        *   `psa_key_derivation_verify_bytes()` --- if each input was either a direct input or a key with usage flag `PSA_KEY_USAGE_VERIFY_DERIVATION`.
+        *   `psa_key_derivation_verify_bytes()` --- if each input was either a direct input, a key with usage flag `PSA_KEY_USAGE_DERIVE`, or a key with usage flag `PSA_KEY_USAGE_VERIFY_DERIVATION`.
         *   `psa_key_derivation_verify_key()` --- under the same conditions as `psa_key_derivation_verify_bytes()`.
 
     If this function returns an error status, the operation enters an error state and must be aborted by calling `psa_key_derivation_abort()`.
@@ -935,7 +935,7 @@ Key derivation functions
         Success.
         The output of the key derivation operation matches ``expected_output``.
     .. retval:: PSA_ERROR_NOT_PERMITTED
-        One of the inputs is a key whose policy does not permit `PSA_KEY_USAGE_VERIFY_DERIVATION`.
+        One of the inputs is a key whose policy permits neither `PSA_KEY_USAGE_DERIVE` nor `PSA_KEY_USAGE_VERIFY_DERIVATION`.
     .. retval:: PSA_ERROR_INVALID_SIGNATURE
         The output of the key derivation operation does not match the value in ``expected_output``.
     .. retval:: PSA_ERROR_INSUFFICIENT_DATA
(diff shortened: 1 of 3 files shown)


doc/crypto/api/ops/kdf.rst
```diff
@@ -906,18 +910,18 @@ Key derivation functions
...
-        Length ot the expected output. This is also the number of bytes that will be read.
+        Length of the expected output. This is also the number of bytes that will be read.
 
     .. return:: psa_status_t
     .. retval:: PSA_SUCCESS
         Success.
         The output of the key derivation operation matches ``expected_output``.
     .. retval:: PSA_ERROR_NOT_PERMITTED
-        One of the inputs is a key whose policy does not permit `PSA_KEY_USAGE_VERIFY_DERIVATION`.
+        One of the inputs is a key whose policy permits neither `PSA_KEY_USAGE_DERIVE` nor `PSA_KEY_USAGE_VERIFY_DERIVATION`.
```
silabs-Kusumit, line 920: If I understand this correctly, the key given as input to `psa_key_derivation_input_key` must have policy as `PSA_KEY_USAGE_DERIVE` or `PSA_KEY_USAGE_VERIFY_DERIVATION`. Is this correct?

gilles-peskine-arm: Correct.

gilles-peskine-arm (approved):
LGTM

athoelke:
Rebased and merged...

PR #120, opened 2023-11-03
Miscellaneous SPAKE2+ support APIs

athoelke:
For SPAKE2+, we need some additions to the PAKE API (also see #73):

* An API to add context data during the operation setup
* A cipher suite attribute for a MAC algorithm (used during key confirmation)
* A PAKE step for key confirmation

### DIFF
--- a/doc/ext-pake/api.db/psa/crypto-pake.h
+++ b/doc/ext-pake/api.db/psa/crypto-pake.h
@@ -52,6 +52,9 @@ psa_status_t psa_pake_output(psa_pake_operation_t *operation,
                              uint8_t *output,
                              size_t output_size,
                              size_t *output_length);
+psa_status_t psa_pake_set_context(psa_pake_operation_t *operation,
+                                  const uint8_t *context,
+                                  size_t context_len);
 psa_status_t psa_pake_set_peer(psa_pake_operation_t *operation,
                                const uint8_t *peer_id,
                                size_t peer_id_len);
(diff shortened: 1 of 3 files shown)


athoelke:

Note some differences to #73, and other fallout:
* In #73, the proposal adds multiple new step types for SPAKE2+. We already have a KEY_SHARE step for the first output/input of SPAKE2+, so this PR only adds a CONFIRM step for the second output/input.
* If we want to support the use of CMAC-AES-128 as the SPAKE2+ MAC (as described in RFC 9383 cipher suites), then we will also need to add the MAC to the list of parameters used in the PAKE input/output SIZE macros: See #116

athoelke:
> * If we want to support the use of CMAC-AES-128 as the SPAKE2+ MAC (as described in RFC 9383 cipher suites), then we will also need to add the MAC to the list of parameters used in the PAKE input/output SIZE macros: See [Add a hash algorithm parameter to the PAKE input and output size macros #116](https://github.com/ARM-software/psa-api/pull/116)

Note: the size of the output for SPAKE2+ depends on the primitive (for the key share step), and the MAC (for the confirmation step). It does not require the hash algorithm. However, other PAKE protocols (e.g SRP) do not use a MAC, and use the output of the cipher-suite hash algorithm for the confirmation step.

silabs-Kusumit:
LGTM

athoelke:
Please look at #124 - if this approach looks better than #116, then we can remove the MAC attribute of the cipher suite (commit https://github.com/ARM-software/psa-api/pull/120/commits/6db9942ae94516b87cdb3466526d1db82391e5af above), and incorporate that into the SPAKE2+ algorithm identifiers instead.

silabs-Kusumit:
> Please look at #124 - if this approach looks better than #116, then we can remove the MAC attribute of the cipher suite (commit [6db9942](https://github.com/ARM-software/psa-api/commit/6db9942ae94516b87cdb3466526d1db82391e5af) above), and incorporate that into the SPAKE2+ algorithm identifiers instead.

Approach in #124 looks better and output size macros requiring MAC for SPAKE2+ is also resolved. We can remove MAC from the ciphersuite.

athoelke:
Rebased, and removed the unnecessary MAC attribute of the cipher-suite. MAC parameterization of a PAKE cipher-suite will be part of the algorithm identifier.

PR #124, opened 2023-11-13
Move PAKE hash parameter to the PAKE algorithm identifier

### LINKED ISSUE #88: PAKE SIZE macros need more arguments

athoelke:
This is alternate approach to #116, in response to issue #88.

* Moved the hash algorithm parameter for a PAKE cipher suite into the PAKE algorithm identifier, instead of storing a separate attribute of the cipher suite.
  - Updated the PAKE algorithm encoding
  - Updated the J-PAKE algorithm macro, and added `PSA_ALG_IS_JPAKE()`
  - Removed the cipher suite accessors for the hash algorithm
* This also makes the hash algorithm value available to the `PSA_PAKE_OUTPUT_SIZE()` and `PSA_PAKE_INPUT_SIZE()` macros.

Fixes #88


athoelke:
Feedback appreciated from @silabs-Kusumit, @silabs-hannes, @oberon-sk, and @yanesca.

silabs-Kusumit:
I think this is a better solution to #88 than #116. 

LGTM

PR #131, opened 2023-12-11
Use PSA Attestation Token as the format for initial attestation reports

### LINKED ISSUE #7: Use the PSA attestation token format in the Attestation API
The attestation token format is currently being standardized as the PSA Attestation token in the draft [datatracker.ietf.org/doc/draft-tschofenig-rats-psa-token](https://datatracker.ietf.org/doc/draft-tschofenig-rats-psa-token) [PSATOKEN] specification.

This has evolved slightly since the v1.0 specification, and now uses allocated claim ids, instead of claim ids from the private use range. The token format described in the v1.0 API should be deprecated in favor of the emerging standard, and the Attestation API needs to be updated with a new version to reference the new format.

There are some open issues and options with how this progresses:

* An implementation that produces a token compliant with [PSATOKEN] is not compliant with the v1.0 Attestation API, so this would probably require an updated specification to be v2.0. But the API is unchanged, so perhaps v1.1 is appropriate?
* Should the legacy v1.0 format be retained in the new version of the specification, perhaps as an appendix, or just removed entirely? It will still be documented in the v1.0 specification.
* Should the new version of the specification provide the same level of detail about the token format as the current specification, provide no details and just refer to the IETF document, or some in-between level of description?
* How should the Attestation API update and IETF document timeline work:
  - Should the new version wait until the IETF document is finalised?
  - Should a beta version of a new Attestation API be published, until the IETF document is finalised?
  - Should a final version of a new Attestation API be published, noting that the IETF document is not yet final?

athoelke:
Update the API to use the PSA attestation token, defined in https://datatracker.ietf.org/doc/draft-tschofenig-rats-psa-token.

The token and report format, CDDL definition, and example token are no longer required in this specification - these are all provided by the PSA Attestation Token specification.

Fixes #7


athoelke:
I would welcome feedback from @thomas-fossati @mathias-arm @SimonFrost-Arm @adrianlshaw @hannestschofenig

thomas-fossati:
> I would welcome feedback from @thomas-fossati @mathias-arm @SimonFrost-Arm @adrianlshaw @hannestschofenig

@athoelke, thanks for the heads up!

> * Is a reference to the external token/report format sufficient?

Yes, I think so.  Especially when the draft becomes an RFC (which should be "soon"), the referenced document will have very strong guarantees in terms of stability.  RFCs are not live documents and cannot move one bit (except for Errata) once they are given a number in the series.

> * Are there specific claim-construction guidelines that should be included in this document, where not provided in the PSA Token specification itself (PSM is too vague to be of any value)?

From the point of view of the API user, the only claim they can control is the `eat_nonce`.  Apart from [§4.1.1](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-4.1.1) and [§5.1.2](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-5.1.2), the recommendations for `eat_nonce` are the same as [§4.1 of EAT](https://www.ietf.org/archive/id/draft-ietf-rats-eat-22.html#section-4.1).  In particular, _"An EAT nonce MUST have at least 64 bits of entropy."_.
Note that the draft makes no recommendations about using `nonce` chained constructions (as mentioned in the PSM).  This may be an area worth elaborating on in the API doc.

> * How much direction should we provide to the legacy format (in the v1.0 specification)?

Backwards compatibility is described in [§4.6](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-4.6).  The gist is:
* attesters should move to the new profile
* verifiers will take care of absorbing legacy

> * Is it worth describing the differences between the legacy and new formats? - or is §4.6 Backwards Compatibility Considerations in the PSA Token spec adequate?

A pointer to that section should be enough.

doc/attestation/overview/report.rst
```diff
@@ -1,272 +1,35 @@
...
 
-This section begins with a description of the information model for the report and then describes the expected format.
+The attestation report returned by the |API| is formatted and encoded as a signed PSA Attestation Token. This is defined in :cite-title:`PSATOKEN`.
 
-Information model
------------------
+The PSA Attestation Token is an incompatible evolution of the original attestation format, that was specified in version 1.0 of the |API|. :issue:`Should this provide a citation for v1.0, or a URL to the 'all-versions' page for the API here?`
 
-The following table describes the mandatory and optional claims in the report:
+To comply with version |docversion| of the |API|, an implementation must only produce attestation reports that conform to :cite:`PSATOKEN`. :issue:`Is this repeat (of the first sentence) necessary or helpful?`
```
SimonFrost-Arm, line 13: I think the repeated warning is appropriate.

SimonFrost-Arm (commented):
I agree with Thomas' replies to the general queries. I've made a couple of responses to the inline Issues.

athoelke:
> > * Are there specific claim-construction guidelines that should be included in this document, where not provided in the PSA Token specification itself (PSM is too vague to be of any value)?
> 
> From the point of view of the API user, the only claim they can control is `eat_nonce`. Apart from [§4.1.1](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-4.1.1) and [§5.1.2](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-5.1.2), the recommendations for `eat_nonce` are the same as [§4.1 of EAT](https://www.ietf.org/archive/id/draft-ietf-rats-eat-22.html#section-4.1). In particular, _"An EAT nonce MUST have at least 64 bits of entropy."_. Note that the draft makes no recommendations about using `nonce` in chained constructions (as mentioned in the PSM). This may be an area worth elaborating on in the API doc.

Note that the API specification is important for implementers as well as users of the API.

The v1.0 spec stated that the Instance ID claim must be constructed as `H(IAK)` for asymmetric IAK, and `H(H(IAK))` for symmetric IAK (using a justification that is not immediately obvious from the citation). This is information is not repeated in the PSA Attestation Token spec, which only describes the `H(IAK)` construction, and PSM has nothing of value to add.

My question regarding claim construction guidance, is partly triggered by this specific description of the Instance ID claim. Is the additional info in v1.0 (choose one):

* Still correct: to be documented in PSA Attestation Token spec?
* Still correct: to be documented in PSA Attestation API (this spec)?
* No longer valid: drop from this spec?
* Not important: does not justify providing different/additional information about the attestation token content/format in this spec?

thomas-fossati:
> > > * Are there specific claim-construction guidelines that should be included in this document, where not provided in the PSA Token specification itself (PSM is too vague to be of any value)?
> > 
> > 
> > From the point of view of the API user, the only claim they can control is `eat_nonce`. Apart from [§4.1.1](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-4.1.1) and [§5.1.2](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-5.1.2), the recommendations for `eat_nonce` are the same as [§4.1 of EAT](https://www.ietf.org/archive/id/draft-ietf-rats-eat-22.html#section-4.1). In particular, _"An EAT nonce MUST have at least 64 bits of entropy."_. Note that the draft makes no recommendations about using `nonce` in chained constructions (as mentioned in the PSM). This may be an area worth elaborating on in the API doc.
> 
> Note that the API specification is important for implementers as well as users of the API.

Ah, thanks for the clarification.  I thought this was only for the user side.

> My question regarding claim construction guidance, is partly triggered by this specific description of the Instance ID claim. Is the additional info in v1.0 (choose one):
> 
> * [ ] Still correct: to be documented in PSA Attestation Token spec?
> * [x] Still correct: to be documented in PSA Attestation API (this spec)?
> * [ ] No longer valid: drop from this spec?
> * [ ] Not important: does not justify providing different/additional information about the attestation token content/format in this spec?

The double-hash construction of the ImplID for symmetric keys is still valid and should be documented here.

athoelke:
> The double-hash construction of the ImplID for symmetric keys is still valid and should be documented here.

Thank you. The follow-up to this specific query, is the generalised question that I originally posed (but with insufficient context to make clear):

* Are there [other] specific claim-construction guidelines that should be included in this document, where not provided in the PSA Token specification itself? - perhaps only relevant to the use of the token within this API?

thomas-fossati:
> > The double-hash construction of the ImplID for symmetric keys is still valid and should be documented here.
> 
> Thank you. The follow-up to this specific query, is the generalised question that I originally posed (but with insufficient context to make clear):
> 
> * Are there [other] specific claim-construction guidelines that should be included in this document, where not provided in the PSA Token specification itself? - perhaps only relevant to the use of the token within this API?

It seems that a major concern for API users is how to _securely_ use the challenge/nonce in the IAT to create chained attestation schemes. To help those who need to build the "Attestation End-Point (AEP)", it would be helpful to convert Section 7.3 of the PSM into actionable information.

----

One thing that I forgot to mention previously is that users may be interested in checking the [security & privacy considerations](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-9) of the PSA token document.  In particular: making sure they trust the challenger, and knowing what potentially privacy-sensitive info they are exposing by using the IAT.

athoelke:
> One thing that I forgot to mention previously is that users may be interested in checking the [security & privacy considerations](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-19.html#section-9) of the PSA token document. In particular: making sure they trust the challenger, and knowing what potentially privacy-sensitive info they are exposing by using the IAT.

I think that this information could be part of the SRA? (see #132)

athoelke:
> It seems that a major concern for API users is how to _securely_ use the challenge/nonce in the IAT to create chained attestation schemes. To help those who need to build the "Attestation End-Point (AEP)", it would be helpful to convert Section 7.3 of the PSM into actionable information.

The use cases section currently describes more of the high level ideas around attestation. Not sure if it would be better if this is more aligned with the discussion of the context in the PSA Attestation Token spec (or just defer to that description?)?

I can see that providing more guidance on how the initial attestation API can be used as part of an attestation protocol, including chained schemes, could be beneficial here.

athoelke:
> The double-hash construction of the ImplID for symmetric keys is still valid and should be documented here.

Can anyone shed light on exactly what the second+third sentence actually means in:

> When using a symmetric key for the IAK, it is recommended that the Instance ID is a double hash of the key - `InstanceID = H(H(IAK))`. This eliminates risks when exposing the key to different HMAC block size. For further information, read RFC 2104.

RFC 2104 is the definition of HMAC. But this construction of Instance ID is not using HMAC.

thomas-fossati:
> Can anyone shed light on exactly what the second+third sentence actually means in:
> 
> > When using a symmetric key for the IAK, it is recommended that the Instance ID is a double hash of the key - `InstanceID = H(H(IAK))`. This eliminates risks when exposing the key to different HMAC block size. For further information, read RFC 2104.
> 
> RFC 2104 is the definition of HMAC. But this construction of Instance ID is not using HMAC.

The thinking behind the double hashing was captured [here](https://trustedfirmware-m.readthedocs.io/en/latest/design_docs/services/symmetric_initial_attest.html?highlight=hmac#instance-id-calculation).

mathias-arm (approved):

doc/attestation/overview/report.rst
```diff
@@ -1,272 +1,39 @@
...
-      -  Byte string (the type byte should be set to ``0x01``. The type byte is described in the `[EAT]` draft.)
+         .. rationale::
+           
+            According to :rfc-title:`2104`, if a HMAC key is longer than the HMAC block size, the key will be first hashed. The hash output is used as the key in HMAC computation.
 
-   *  -  -75010
-      -  Verification service indicator
-      -  ``arm_psa_origination``
-      -  Text string
+            When HMAC is used to authenticate the token, and IAK is longer than the HMAC block size, then ``HMAC(IAK, token) == HMAC(H(IAK), token)``. If Instance ID is defined to be ``H(IAK)``, then an attacker can then use the Instance ID value in an attestation token to fake malicious reports by using Instance ID as the HMAC key.
```
thomas-fossati, line 37: removed double "then"

```suggestion
            When HMAC is used to authenticate the token, and IAK is longer than the HMAC block size, then ``HMAC(IAK, token) == HMAC(H(IAK), token)``. If Instance ID is defined to be ``H(IAK)``, then an attacker can use the Instance ID value in an attestation token to fake malicious reports by using Instance ID as the HMAC key.
```

SimonFrost-Arm (approved):

athoelke:
Oops. Broken rebase was force-pushed... but recovered now.

PR #138, opened 2024-01-08
PSA_ERROR_INVALID_HANDLE: clarify that it's about existing objects

gilles-peskine-arm:
The current wording makes it look like the error also applies when creating an object with a caller-chosen handle and the caller's choice is not acceptable. See https://lists.trustedfirmware.org/archives/list/mbed-tls@lists.trustedfirmware.org/thread/DOBLFTSJMN2BGW3E7J6HVX6CXZ4YBKK3/

### DIFF
--- a/doc/status-code/api/status-codes.rst
+++ b/doc/status-code/api/status-codes.rst
@@ -205,7 +205,9 @@ Error codes
 
    .. summary:: A status code that indicates that a handle parameter is not valid.
 
-   A function can return this error any time a handle parameter is invalid.
+   A function can return this error any time it expects a parameter to be a handle to an existing object, but the handle is invalid.
+
+   This status code is not intended for cases where the parameter is a value that is chosen by the caller, but the value does not satisfy a validity condition for handles of this kind. In such cases, it is recommended that a function returns `PSA_ERROR_ALREADY_EXISTS` if value is valid but already in use, and `PSA_ERROR_INVALID_ARGUMENT` or a more specific status code if the value is otherwise invalid.
 
 
 .. macro:: PSA_ERROR_BAD_STATE
--- a/doc/status-code/api/status-codes.rst
+++ b/doc/status-code/api/status-codes.rst
@@ -205,7 +205,7 @@ Error codes
 
    .. summary:: A status code that indicates that a handle parameter is not valid.
 
-   A function can return this error any time it expects a parameter to be a handle to an existing object, but the handle is invalid.
+   A function can return this error any time it expects a parameter to be a handle to an existing resource, but the handle is invalid.
 
    This status code is not intended for cases where the parameter is a value that is chosen by the caller, but the value does not satisfy a validity condition for handles of this kind. In such cases, it is recommended that a function returns `PSA_ERROR_ALREADY_EXISTS` if value is valid but already in use, and `PSA_ERROR_INVALID_ARGUMENT` or a more specific status code if the value is otherwise invalid.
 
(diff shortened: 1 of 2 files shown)


MarcusJGStreets:
If you make this change, what would be the difference between PSA_ERROR_INVALID_HANDLE and PSA_ERROR_DOES_NOT_EXIST?

What do you return if the value is out of range? PSA_KEY_ID_USER_MIN to PSAKEY_ID_USER_MAX or PSA_KEY_ID_VENDOR_MIN to PSA_KEY_ID_VENDOR_MAX?

gilles-peskine-arm:
This is intended only to clarify the existing specification, not as a change of behavior.

Crypto API functions don't return `PSA_ERROR_DOES_NOT_EXIST`. Storage does, and I think storage doesn't use `PSA_ERROR_INVALID_HANDLE`.

> What do you return if the value is out of range?

When attempting to use an existing key: `INVALID_HANDLE`, same as if the value is in range but the key does not exist. Note that a key can exist outside the ranges you specify, if it's a vendor-specific built-in/magic key.

When attempting to create a key: `INVALID_ARGUMENT`, same as other arguments with a value that isn't acceptable.

athoelke:
> If you make this change, what would be the difference between `PSA_ERROR_INVALID_HANDLE` and `PSA_ERROR_DOES_NOT_EXIST`?

The use of `PSA_ERROR_INVALID_HANDLE` stems from a legacy. The initial development of this API had distinct _handles_ for using a key, and _identifiers_ for naming a persistent key. Just prior to 1.0.0, we merged the concepts and used key ids for everything - but we didn't change the API of all key-using functions to return `PSA_ERROR_DOES_NOT_EXIST` instead of `PSA_ERROR_INVALID_HANDLE` for key ids that did not match any key.

If starting from scratch, I would use `PSA_ERROR_DOES_NOT_EXIST`, as for other identifiers, such as in the Storage and Firmware Update APIs. However, changing this aspect of the API would be a breaking change so we live with `PSA_ERROR_INVALID_HANDLE`.

doc/status-code/api/status-codes.rst
```diff
@@ -205,7 +205,9 @@ Error codes
 
    .. summary:: A status code that indicates that a handle parameter is not valid.
 
-   A function can return this error any time a handle parameter is invalid.
+   A function can return this error any time it expects a parameter to be a handle to an existing object, but the handle is invalid.
```
athoelke, line 208: Prefer 'resource' to 'object' - consistent with the other error code descriptions.

doc/status-code/api/status-codes.rst
```diff
@@ -205,7 +205,9 @@ Error codes
...
+
+   This status code is not intended for cases where the parameter is a value that is chosen by the caller, but the value does not satisfy a validity condition for handles of this kind. In such cases, it is recommended that a function returns `PSA_ERROR_ALREADY_EXISTS` if value is valid but already in use, and `PSA_ERROR_INVALID_ARGUMENT` or a more specific status code if the value is otherwise invalid.
```
athoelke, line 210: With the change to the existing paragraph, I am not convinced that this is needed? - the new scoping in para 1 clearly indicates that this return code is only applicable for APIs where the 'handle' parameter is meant to refer to an existing item.

APIs in which the caller is requesting a specific value for a new handle are not in that scope.

If we do want to keep this, I find the wording of the first sentence unclear. Would this be any better:

> This status code is not recommended when a caller is requesting a specific handle value to be used for a new handle.

athoelke (changes requested):
Overall, I think this is valuable clarification. Some comments on the wording.

This will result in a patch release for the Status code API as well as the Crypto API.

athoelke (approved):
LGTM

PR #139, opened 2024-01-09
Minor fixes 

MarcusJGStreets:


athoelke (approved):
Thank you for fixing the mitigation description in the final Threat.

I will manage updates to copyright dates when finalising the document for publication, so no need to update this PR

PR #142, opened 2024-01-10
Clarify the required behavior of concurrent calls

### LINKED ISSUE #134: Crypto: soften the requirement for concurrency
According to [version 1.1 of the crypto API specification](https://arm-software.github.io/psa-api/crypto/1.1/overview/conventions.html#concurrent-calls):

> Concurrent calls are performed correctly, as if the calls were executed in sequence, provided that [not important here]

The intent is that this only holds for functional correctness. For example, at any given point in time, a key identifier either exists or doesn't exist, there isn't some other state where querying the key identifier says it doesn't exist but attempting to create it returns `ALREADY_EXISTS`. But this doesn't hold for all observations. Specifically, it is possible, and [generally desirable](https://github.com/Mbed-TLS/mbedtls/pull/8618#discussion_r1430640587), that library calls are not linearizable with respect to resource exhaustion. For example, it should be permitted that, starting from a given system state, with one thread calling `f1()` and another thread calling `f2()`:

* both calls succeed, with `f1()` taking place before `f2()` (if the order is observable);
* both calls succeed, with `f2()` taking place before `f1()` (if the order is observable);
* At least one of the calls in `f1() || f2()` fails due to some resource exhaustion (e.g. `PSA_ERROR_INSUFFICIENT_MEMORY`, `PSA_ERROR_INSUFFICIENT_STORAGE`, `PSA_ERROR_COMMUNICATION_FAILURE`).

This is in keeping with the general intent of the crypto API specification, which never makes any guarantee of availability. But it violates the letter of the statement quoted above.

athoelke:
Relax/clarify the statement regarding the required behavior of implementations that support concurrent execution of calls to the API.

Fixes #134

### DIFF
--- a/doc/crypto/overview/conventions.rst
+++ b/doc/crypto/overview/conventions.rst
@@ -393,7 +393,7 @@ calls to the API whose execution can overlap in time.
 
     The consistency requirement does not apply to errors that arise
     from resource failures or limitations. For example, errors resulting from
-    resource constraints can arise in concurrent execution that do not arise in
+    resource exhaustion can arise in concurrent execution that do not arise in
     sequential execution.
 
     As an example of this rule: suppose two calls are executed concurrently
--- a/doc/crypto/overview/conventions.rst
+++ b/doc/crypto/overview/conventions.rst
@@ -370,37 +370,45 @@ In some environments, an application can make calls to the |API| in
 separate threads. In such an environment, *concurrent calls* are two or more
 calls to the API whose execution can overlap in time.
 
-The result of two or more concurrent calls must be consistent with the same set
-of calls being executed sequentially in some order, provided that they obey the
-following constraints:
+**Sequential consistency**
+    The result of two or more concurrent calls must be consistent with the
+    same set of calls being executed sequentially in some order, provided that
+    the calls obey the following constraints:
 
-*   There is no overlap between an output parameter of one call and an input or
-    output parameter of another call. Overlap between input parameters is
-    permitted.
+    *   There is no overlap between an output parameter of one call and an
+        input or output parameter of another call. Overlap between input
+        parameters is permitted.
 
-*   A call to destroy a key must not overlap with a concurrent call to any of
-    the following functions:
+    *   A call to :code:`psa_destroy_key()` must not overlap with a concurrent
+        call to any of the following functions:
 
-    *   Any call where the same key identifier is a parameter to the call.
-    *   Any call in a multi-part operation, where the same key identifier was
-        used as a parameter to a previous step in the multi-part operation.
+        -   Any call where the same key identifier is a parameter to the call.
+        -   Any call in a multi-part operation, where the same key identifier
+            was used as a parameter to a previous step in the multi-part
+            operation.
 
-*   Concurrent calls must not use the same operation object.
+    *   Concurrent calls must not use the same operation object.
 
-If any of these constraints are violated, the behavior is undefined.
+    If any of these constraints are violated, the behavior is undefined.
 
-For example, suppose two calls are executed concurrently which both attempt to
-create a new key with the same key identifier that is not already in the key
-store. Then:
+    Note that this consistency requirement does not apply to errors that arise
+    from resource failures or limitations, for example, errors resulting from
+    resource constraints can arise in concurrent execution that do not arise in
+    sequential execution.
 
-*   If one call returns :code:`PSA_ERROR_ALREADY_EXISTS`, then the other call
-    must succeed.
-*   If one of the calls succeeds, then the other must fail: either with
-    :code:`PSA_ERROR_ALREADY_EXISTS` or some other error status.
-*   Both calls can fail with error codes that are not
-    :code:`PSA_ERROR_ALREADY_EXISTS`.
+    For example, suppose two calls are executed concurrently which both attempt
+    to create a new key with the same key identifier that is not already in the
+    key store. Then:
 
-If the application concurrently modifies an input parameter while a function
-call is in progress, the behavior is undefined.
+    *   If one call returns :code:`PSA_ERROR_ALREADY_EXISTS`, then the other
+        call must succeed.
+    *   If one of the calls succeeds, then the other must fail: either with
+        :code:`PSA_ERROR_ALREADY_EXISTS` or some other error status.
+    *   Both calls can fail with error codes that are not
+        :code:`PSA_ERROR_ALREADY_EXISTS`.
+
+**Parameter stability**
+    If the application concurrently modifies an input parameter while a
+    function call is in progress, the behavior is undefined.
 
 Individual implementations can provide additional guarantees.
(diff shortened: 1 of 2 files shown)


doc/crypto/overview/conventions.rst
```diff
@@ -370,8 +370,9 @@ In some environments, an application can make calls to the |API| in
 separate threads. In such an environment, *concurrent calls* are two or more
 calls to the API whose execution can overlap in time.
 
-Concurrent calls are performed correctly, as if the calls were executed in
-sequence, provided that they obey the following constraints:
+The result of two or more concurrent calls must be consistent with the same set
+of calls being executed sequentially in some order, provided that they obey the
```
Ryan-Everett-arm, line 374: This new statement is definitely an improvement. However, we want to specifically allow situations in which a call fails with a resource-exhaustion error where it is impossible for such a failure to occur in sequential execution. There is an example of this scenario here https://github.com/Mbed-TLS/mbedtls/pull/8618#issuecomment-1858264256. My interpretation of the suggested statement (there exists some valid ordering of the calls where for all `k`: [call `k` succeeds in the sequential execution] iff [call `k` succeeds in the concurrent execution]) doesn't permit this behaviour.

It is difficult to word the desired behaviour concisely. One suggestion is to modify the statement to say "The result (excluding resource exhaustion errors) of two or more..." and then add some text after the constraint list saying something like:
```
It is always permissible for a concurrent call to fail with a
resource exhaustion error (PSA_ERROR_INSUFFICIENT_MEMORY,
PSA_ERROR_INSUFFICIENT_STORAGE, PSA_ERROR_COMMUNICATION_FAILURE).
```

Ryan-Everett-arm (commented):
It was a good idea to add an example here, the intuition of concurrency constraints can be difficult to understand. I'm not sure if the proposed constraint will be interpreted as intended, though; see my comment below.

doc/crypto/overview/conventions.rst
athoelke: I see your point. In some ways I was trying to avoid explicitly complicating the rule by discussing resource exhaustion, in the use of 'consistent with' - which is intentionally not 'the same result as if'. However, that probably depends on a nuanced reading of 'consistent with', which is not ideal for a spec that is not in a reader's primary language.

I would also prefer to avoid getting mired in description of side-effects on state that is explicitly part of the API programming model. (For example, keys in the key store and operation objects). Although that could clarify the intent of the concurrency rule - that failures due to interactions with non-API state (such as memory, storage etc) are not subject to the linearizability requirement - this gets very technical.

In practice - given the existing rules about undefined behavior if a key is destroyed concurrently with a call that uses it, and concurrent use of an operation object - the only API-state changing calls that are in scope are ones that create a key, and their interaction with other key creation or key usage calls on the same key identifier.

So we could either try and improve the clarification of the general rule. Or instead provide explicit guidance for the interaction of key creation and key usage APIs when called concurrently?

athoelke: However, I quite like the idea of trying to note the general caveat that errors resulting from resource constraints can arise in concurrent execution that do not arise in sequential execution.

athoelke: I think the mental 'trick' to understand what matters in the concurrency requirement is that the rule only cares about the behavior of the calls in relation to the specification-defined programming model. Therefore 'consistent with' does not mean 'the same as a sequential execution in the same implementation', but 'consistent with a valid sequential execution of the abstract programmer's model' - the abstract model does not provide or require guarantees about when resources might or might not be exhausted, whereas an implementation can provide resource-exhaustion-free guarantees for some calls.

However, that trick is not obvious to an implementer - they are just thinking about their own implementation.

Ryan-Everett-arm: > I see your point. In some ways I was trying to avoid explicitly complicating the rule by discussing resource exhaustion, in the use of 'consistent with' - which is intentionally not 'the same result as if'. However, that probably depends on a nuanced reading of 'consistent with', which is not ideal for a spec that is not in a reader's primary language.
> 
> I would also prefer to avoid getting mired in description of side-effects on state that is explicitly part of the API programming model. (For example, keys in the key store and operation objects). Although that could clarify the intent of the concurrency rule - that failures due to interactions with non-API state (such as memory, storage etc) are not subject to the linearizability requirement - this gets very technical.
> 
> In practice - given the existing rules about undefined behavior if a key is destroyed concurrently with a call that uses it, and concurrent use of an operation object - the only API-state changing calls that are in scope are ones that create a key, and their interaction with other key creation or key usage calls on the same key identifier.
> 
> So we could either try and improve the clarification of the general rule. Or instead provide explicit guidance for the interaction of key creation and key usage APIs when called concurrently?



I agree that we should avoid discussing key stores and the like.

If we only provide guidance for interactions involving key creation then it is unclear whether we allow non-linearizable resource exhaustion errors in other cases where it is desirable. For example, the usage of persistent keys that are valid but not currently in memory.

In my opinion it is better to be general, while still making clear that non-linearizable resource exhaustion errors are permitted where calls would otherwise succeed.

Ryan-Everett-arm: > I think the mental 'trick' to understand what matters in the concurrency requirement is that the rule only cares about the behavior of the calls in relation to the specification-defined programming model. Therefore 'consistent with' does not mean 'the same as a sequential execution in the same implementation', but 'consistent with a valid sequential execution of the abstract programmer's model' - the abstract model does not provide or require guarantees about when resources might or might not be exhausted, whereas an implementation can provide resource-exhaustion-free guarantees for some calls.
> 
> However, that trick is not obvious to an implementer - they are just thinking about their own implementation.

By looking at the proposed constraints with this perspective I now see how the statement works. Your last sentence is true: as an implementer I first read the constraints from the implementation-fixed perspective you describe.

athoelke: I've update the PR with some organising structure (so it is clear which text relates to the 'consistency' rule), and added the resource-related error caveat/relaxation.

athoelke: Just thought: I should probably remove the 'Note that...' from the caveat, as it makes it appear informative, rather that normative and providing clear permission to an implementer?

Ryan-Everett-arm: I am now happy to approve these changes, I think it now reads clearly and correctly - I agree that removing "Note that" is a good idea.

Ryan-Everett-arm (approved):
LGTM

doc/crypto/overview/conventions.rst
```diff
@@ -370,25 +370,45 @@ In some environments, an application can make calls to the |API| in
...
+            was used as a parameter to a previous step in the multi-part
+            operation.
+
+    *   Concurrent calls must not use the same operation object.
+
+    If any of these constraints are violated, the behavior is undefined.
+
+    The consistency requirement does not apply to errors that arise
+    from resource failures or limitations. For example, errors resulting from
+    resource constraints can arise in concurrent execution that do not arise in
```
gilles-peskine-arm, line 396: Minor: “constraints” is vague, I think “exhaustion” would make the example clearer.

athoelke: Fixed

PR #148, opened 2024-01-16
Adding Iterator and remane functions

MarcusJGStreets:
Added in PS storage initially. 
If thee is need for them in ITS, the same additions can be made there

### DIFF
--- a/doc/storage/api/api.rst
+++ b/doc/storage/api/api.rst
@@ -9,7 +9,7 @@ Status codes
 
 The |API| uses the status code definitions that are shared with the other PSA Certified APIs.
 
-The following elements are defined in :file:`psa/error.h` from :cite-title:`PSA-STAT` (previously defined in :cite:`PSA-FFM`):
+The following elements are defined in :file:`psa/error.h` from :cite-title:`PSA-STAT` (previously defined in :cite:`PSA-FF-M`):
 
 .. code-block:: xref
 
@@ -97,10 +97,33 @@ These definitions must be defined in the header file :file:`psa/storage_common.h
 
     The data associated with the ``uid`` does not require replay protection. This can permit faster storage --- but it permits an attacker with physical access to revert to an earlier version of the data.
 
+.. macro:: PSA_STORAGE_FLAG_OVERWRITE
+    (1u << 3)
+
+    Flag instructing the `psa_ps_rename()` function to replace existing stored data. 
+
 .. macro:: PSA_STORAGE_SUPPORT_SET_EXTENDED
     (1u << 0)
 
     Flag indicating that `psa_ps_create()` and `psa_ps_set_extended()` are supported.
+
+
+.. macro:: PSA_STORAGE_SUPPORT_RENAME
+    (1u << 1)
+
+    Flag indicating that `psa_ps_rename()` is supported.
+
+.. macro:: PSA_STORAGE_SUPPORT_ITERATION
+    (1u << 2)
+
+    Flag indicating that `psa_ps_iterator_init()` and `psa_ps_iterator_next()` are supported.
+
+.. macro:: PSA_STORAGE_ITERATOR_CTX__SIZE
+    :definition: /* implementation-defined value */
+
+    .. summary::
+        A sufficient buffer size for a iterator context, in bytes.
+
 
 .. _ITS-API:
 
@@ -235,9 +258,9 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
 
     * The value of ``data_offset`` must be less than or equal to the length of ``uid``.
 
-    * If ``data_ffset`` is greater than ``uid.size``, no data is retrieved and the functions returns PSA_INVALID_ARGUMENT.
+    * If ``data_offset`` is greater than ``uid.size``, no data is retrieved and the functions returns PSA_INVALID_ARGUMENT.
 
-    * If ``data_size`` is not zero, ``p_data`` must mot be ``NULL``.
+    * If ``data_size`` is not zero, ``p_data`` must not be ``NULL``.
 
     * The call must have access to the memory from ``p_data`` to ``p_data + data_size - 1``.
 
@@ -353,7 +376,7 @@ These definitions must be defined in the header file :file:`psa/protected_storag
 .. function:: psa_ps_set
 
     .. summary::
-        Set the data associated with the specified ``uid``.
+        Set the data associated with the specified ``uid``, replacing any previous data. 
 
     .. param:: psa_storage_uid_t uid
         The identifier for the data.
@@ -379,6 +402,9 @@ These definitions must be defined in the header file :file:`psa/protected_storag
         * The ``uid`` is ``0``.
 
         * The operation failed because caller cannot access some or all of the memory in the range [``p_data``, ``p_data + data_length - 1``].
+        
+        * the uid exists and ``data_length`` is greater then ```capacity`` 
+        
     .. retval:: PSA_ERROR_NOT_SUPPORTED
         The operation failed because one or more of the flags provided in ``create_flags`` is not supported or is not valid.
     .. retval:: PSA_ERROR_INSUFFICIENT_STORAGE
@@ -388,10 +414,13 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     .. retval:: PSA_ERROR_GENERIC_ERROR
         The operation failed because of an unspecified internal failure.
 
-    The newly created asset has a capacity and size that are equal to ``data_length``.
-
-
+    If ``uid`` does not already exist, creates a new asset, the newly created asset has a capacity and size that are equal to ``data_length``.
+    
+    If ``uid`` exists and was not created with  `PSA_STORAGE_FLAG_WRITE_ONCE` and  ``capacity`` is greater to equal to ``data_length`` replaces the existing contents with ``p_data``. ``uid.size`` is set to ``data_length``.
+    
     * The ``uid`` value must not be zero.
+    
+    * If ``uid`` exists ``data_length`` must not be greater than ``uid.capacity``. 
 
     * If ``uid`` exists it must not have been created as with `PSA_STORAGE_FLAG_WRITE_ONCE` - would result in ``PSA_ERROR_NOT_PERMITTED``
 
@@ -454,7 +483,7 @@ These definitions must be defined in the header file :file:`psa/protected_storag
 
     * If ``data_offset`` is greater than ``uid.size`` the function retrieves no data and returns ``PSA_ERROR_INVALID_ARGUMENT``
 
-    * If ``data_size`` is not zero, ``p_data`` must mot be ``NULL``.
+    * If ``data_size`` is not zero, ``p_data`` must not be ``NULL``.
 
     * The call must have access to the memory from ``p_data`` to ``p_data + data_size - 1``.
 
@@ -508,6 +537,7 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     * Even if all parameters are correct, the function can fail in the case of a storage failure.
 
 
+
 .. function:: psa_ps_remove
 
     .. summary::
@@ -540,6 +570,7 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     * If the location ``uid`` exists, it and any metadata are removed.
 
     * Even if all parameters are correct, the function can fail in the case of a storage failure.
+
 
 
 .. function:: psa_ps_create
@@ -594,7 +625,7 @@ These definitions must be defined in the header file :file:`psa/protected_storag
 .. function:: psa_ps_set_extended
 
     .. summary::
-        Overwrite part of the data of the specified ``uid``.
+        Overwrite part of the data of the specified ``uid``, leaving remaining data unchanged. 
 
     .. param:: psa_storage_uid_t uid
         The unique identifier for the asset.
@@ -632,15 +663,15 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     .. retval:: PSA_ERROR_NOT_PERMITTED
         The operation failed because it was attempted on an asset which was written with the flag `PSA_STORAGE_FLAG_WRITE_ONCE`.
 
-    Sets partial data into an asset based on the given identifier, ``data_offset``, ``data length`` and ``p_data``.
+    Sets partial data into an asset based on the given identifier, ``data_offset``, ``data length`` and ``p_data``. 
 
     Before calling this function, the storage must have been reserved with a call to `psa_ps_create()`. It can also be used to overwrite data in an asset that was created with a call to `psa_ps_set()`.
 
     Calling this function with ``data_length == 0`` is permitted. This makes no change to the stored data.
 
     This function can overwrite existing data and/or extend it up to the capacity for the ``uid`` specified in `psa_ps_create()` but cannot create gaps.
-
-    This function is optional. Consult the platform documentation to determine if it is implemented or perform a call to `psa_ps_get_support()`. This function must be implemented if `psa_ps_get_support()` returns `PSA_STORAGE_SUPPORT_SET_EXTENDED`.
+    
+      This function is optional. Consult the platform documentation to determine if it is implemented or perform a call to `psa_ps_get_support()`. This function must be implemented if `psa_ps_get_support()` returns `PSA_STORAGE_SUPPORT_SET_EXTENDED`.
 
     * The ``uid`` value must not be zero.
 
@@ -649,7 +680,7 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     * ``data_offset <= size``
 
     * ``data_offset + data_length <= capacity``
-
+    
     * Even if all parameters are correct, the function can fail in the case of a storage failure.
 
     On Success:
@@ -657,9 +688,136 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     * ``size = max(size, data_offset + data_length)``
 
     * ``capacity`` unchanged.
+    
+    * Data in the ranges 0 to ``data_offset`` is not modified. 
+    
+    * If ``data_offset + data_length < size`` then data in the range ``data_offset + data_length`` to ``size`` is not modified. 
 
 
 
+.. function:: psa_ps_rename
+
+   .. summary::
+        Atomically renames the storage location with the specified ``uid`` to a ``uid_new``.
+
+   .. param:: psa_storage_uid_t uid
+        The current identifier for the data.
+
+   .. param:: psa_storage_uid_t uid_new
+        The new identifier for the data.
+
+   .. param:: psa_storage_rename_flags_t rename_flags
+        The flags must be either ``PSA_STORAGE_FLAG_NONE`` or ``PSA_STORAGE_FLAG_OVERWRITE``
+
+   .. return:: psa_status_t
+        A status indicating the success or failure of the operation.
+
+   .. retval:: PSA_SUCCESS
+        The operation completed successfully.
+
+   .. retval:: PSA_ERROR_ALREADY_EXISTS
+        Storage with the specified ``uid_new`` already exists. 
+
+   .. retval:: PSA_ERROR_GENERIC_ERROR
+        The operation failed because of an unspecified internal failure.
+
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The operation failed because either:
+
+        * ``uid`` is ``0``.
+        * ``uid_new`` is ``0``
+        * the ``psa_storage_rename_flags_t`` has a value set other than ``PSA_STORAGE_FLAG_OVERWRITE``
+
+   .. retval:: PSA_ERROR_NOT_PERMITTED
+        The operation failed because either:                
+        
+        * ``uid_new`` exists and was created with ``PSA_STORAGE_FLAG_WRITE_ONCE``.        
+        * ``uid_new`` exists and ``PSA_STORAGE_FLAG_OVERWRITE`` is not specified
+
+        * ``uid`` was created with ``PSA_STORAGE_FLAG_NO_CONFIDENTIALITY`` and ``uid_new`` exists and does not have that flag set
+        * ``uid`` was created with ``PSA_STORAGE_FLAG_NO_REPLAY_PROTECTION`` and ``uid_new`` exists and does not have that flag set
+
+   .. retval:: PSA_ERROR_NOT_SUPPORTED
+        The implementation does not support the operation.
+
+   .. retval:: PSA_ERROR_STORAGE_FAILURE
+        The operation failed because the physical storage has failed (Fatal error).     
+
+
+   The function renames ``uid`` to ``uid_new`` retaining the storage flags  that ``uid`` was created with. 
+
+   If the caller specifies ``PSA_STORAGE_FLAG_OVERWRITE`` the operation atomically replaces the exisitng contents of ``uid_new`` with those of ``uid``.
+
+   However, ``uid`` must not have been created with more permissive Storage flags than ``uid_new``
+        
+   Except in the case of ``PSA_ERROR_STORAGE_FAILURE``, in which case no guarantees can be made, the operation shall either succeed or leave storage unchanged. 
+
+
+
+
+.. function:: psa_ps_iterator_init
+
+   .. summary:: 
+       Initialises an iterator that can be used to return a list of uids in the Protected Storage .
+
+   .. param::  uint8_t* context
+       A pointer to a context for this iterator. This is set to a new value on success and is undefined on error. The content of the iterator is implementation defined.  The pointer MUST be valid and MUST point to a block of memory of the size given by the ``PSA_STORAGE_ITERATOR_CTX__SIZE`` macro to which the caller has write access. 
+   .. param:: psa_storage_uid_t mask
+       A value used to filter the results included in this iteration. 
+   .. param:: psa_storage_uid_t filter
+       A value used to filter the results included in this iteration. 
+
+   .. return:: psa_status_t
+       A status indicating the success or failure of the operation.
+
+   .. retval:: PSA_SUCCESS
+       The operation completed successfully.
+   .. retval:: PSA_ERROR_STORAGE_FAILURE
+       The operation failed because the physical storage has failed (Fatal error).
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+       The operation failed because the caller cannot access some or all of the memory in the range [``context``, ``context + PSA_STORAGE_ITERATOR_CTX__SIZE - 1``].    
+
+
+   Any bit positions where the mask is set to one must match the value in the filter.
+   
+   That is, the iterator returns those values where ``uid & mask = filter``
+
+   The iterator will only returns those ``uid`` that were created by the caller. It MUST not return any ``uid`` created by a different user. 
+
+   An iterator is not required to return uids in any specific order, but MUST return them in a consistent order each time it is called. For example, if an implementation returns entries in numerical order, it should not arbitrarily change to returning them in creation order. However, the caller should not make assumptions as to the order in which entries are returned, except that each uid will be returned only once in each iteration.
+
+   Changes to storage by other users MUST NOT affect any open iterations. 
+
+   A caller may initialize multiple iteration contexts at the same time. Each iteration shall be independent. Calling ``psa_ps_iterator_next()`` on one iterator MUST not effect any other open iteration. 
+
+
+.. function:: psa_ps_iterator_next
+   .. summary::
+
+   Returns the next ``uid`` in this iteration. 
+
+   .. param::  uint8_t* context
+       A pointer to a context for this iterator as returned by ``psa_ps_iterator_init()``.  The content of the iterator will change on success and is undefined on error.
+
+   .. param:: *psa_storage_uid_t result
+        A pointer to the location in which to store ``uid``. 
+
+   .. return:: psa_status_t
+       A status indicating the success or failure of the operation.
+
+   .. retval:: PSA_SUCCESS
+       The operation completed successfully.
+   .. retval:: PSA_ERROR_EOF
+       The iterator has returned all the uids that match this iteration.
+   .. retval:: PSA_ERROR_STORAGE_FAILURE
+       The operation failed because the physical storage has failed (Fatal error).
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+       The operation failed because either:
+       
+       * The provided context is not valid. 
+       
+       * The caller cannot access the memory at ``result``
+       
 .. function:: psa_ps_get_support
 
     .. return:: uint32_t
@@ -670,3 +828,5 @@ These definitions must be defined in the header file :file:`psa/protected_storag
     Currently defined flags are limited to:
 
     * `PSA_STORAGE_SUPPORT_SET_EXTENDED`
+    * `PSA_STORAGE_SUPPORT_RENAME`
+    * `PSA_STORAGE_SUPPORT_ITERATION`
--- a/doc/storage/about.rst
+++ b/doc/storage/about.rst
@@ -36,10 +36,16 @@
     Documentation clarifications.
 
 .. release:: 1.0.3
-    :date: ? 2024
+    :date: August 2023
     :confidentiality: Non-confidential
 
-    Provide a Security Risk Assessment.
+    Added draft SRA.
+
+.. release:: 1.1.0
+    :date: January 2024
+    :confidentiality: Non-confidential
+
+    Added rename and iteration 
 
 .. release-info::
     :extend:
@@ -64,16 +70,11 @@
     :doc_no: ARM IHI 0097
     :url: arm-software.github.io/psa-api/status-code
 
-.. reference:: PSA-FFM
+.. reference:: PSA-FF-M
     :title: Arm® Platform Security Architecture Firmware Framework
     :doc_no: ARM DEN 0063
-    :url: developer.arm.com/documentation/den0063
+    :url: pages.arm.com/psa-apis
 
-.. reference:: SP800-30
-    :title: NIST Special Publication 800-30 Revision 1: Guide for Conducting Risk Assessments
-    :author: NIST
-    :publication: September 2012
-    :url: doi.org/10.6028/NIST.SP.800-30r1
 
 .. Glossary terms used in this specification
 


athoelke (changes requested):
Need to clarify the conceptual behavior of `psa_ps_rename()` - 'replace-item' or 'move-content'?

doc/storage/api/api.rst
```diff
@@ -379,6 +402,9 @@ These definitions must be defined in the header file :file:`psa/protected_storag
         * The ``uid`` is ``0``.
 
         * The operation failed because caller cannot access some or all of the memory in the range [``p_data``, ``p_data + data_length - 1``].
+        
+        * the uid exists and ``data_length`` is greater then ```capacity``
```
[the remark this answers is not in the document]
MarcusJGStreets: I think the intention was always that if a file was created with capacidy = x then any set with size < x will succeed. 

I thought the initial intention was that if size > capacity - for either set or set_extended - then it fials. 

If the intention was that it might succeed, but cannot be guaranteed, that is also fine.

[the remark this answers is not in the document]
gilles-peskine-arm: > I think the intention was always that if a file was created with capacity = x then any set with size < x will succeed.

I don't know what the intent was, but I find this requirement surprising. In a log-structured filesystem, overwriting a file with less content takes some disk space. Even deleting a file can require some disk space in some filesystems.

doc/storage/api/api.rst
```diff
@@ -649,17 +680,144 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   
+   That is, the iterator returns those values where ``uid & mask = filter``
+
+   The iterator will only returns those ``uid`` that were created by the caller. It MUST not return any ``uid`` created by a different user. 
+
+   An iterator is not required to return uids in any specific order, but MUST return them in a consistent order each time it is called. For example, if an implementation returns entries in numerical order, it should not arbitrarily change to returning them in creation order. However, the caller should not make assumptions as to the order in which entries are returned, except that each uid will be returned only once in each iteration.
+
+   Changes to storage by other users MUST NOT affect any open iterations. 
+
+   A caller may initialize multiple iteration contexts at the same time. Each iteration shall be independent. Calling ``psa_ps_iterator_next()`` on one iterator MUST not effect any other open iteration.
```
athoelke, line 791: This is a critical part of the specification - combined with the absence of a 'resource release' function, it indicates [implicitly] that there cannot be any RoT service-side state associated with an iterator, without causing an unmanageable resource exhaustion problem.

doc/storage/api/api.rst
```diff
@@ -649,17 +680,144 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   If the caller specifies ``PSA_STORAGE_FLAG_OVERWRITE`` the operation atomically replaces the exisitng contents of ``uid_new`` with those of ``uid``.
+
+   However, ``uid`` must not have been created with more permissive Storage flags than ``uid_new``
+        
+   Except in the case of ``PSA_ERROR_STORAGE_FAILURE``, in which case no guarantees can be made, the operation shall either succeed or leave storage unchanged. 
+
+
+
+
+.. function:: psa_ps_iterator_init
```
athoelke, line 758: The iteration API needs a 'usage' example to show how applications are expected to use the interfaces, and what kinds of operation can be done while iterating.

It must also be very clearly documented that there is no 'close'/'resource release' function for iterators
I think this aspect of the API design should be much more clearly stated, as it has a critical effect on the way applications use them, and on the possible implementation choices. In particular, a RoT service implementation cannot afford to have service-side state relating to an iterator, given that:

* An application is permitted to have more than one active iterator
* It is not assumed or required that the application always operates an iterator (calls 'next' on it) until it reaches the end of the iteration

I think this interface design is acceptable, and implementations can maintain adequate iterator state in the iterator context (in the application) to meet the requirements of the use cases for the API.

athoelke (changes requested):
The iterator API needs:

* Clarification about where the iteration state is managed (the API design partly dictates this, but is not explicit)
* Would benefit from a usage example
* Needs clarity on interleaved update operations
* Should probably use an implementation-defined type, rather than a byte-array for the context

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+        * ``uid`` is ``0``.
+        * ``uid_new`` is ``0``
+        * the ``psa_storage_rename_flags_t`` has a value set other than ``PSA_STORAGE_FLAG_REPLACE``
+
+   .. retval:: PSA_ERROR_NOT_PERMITTED
+        The operation failed because either:                
+        
+        * ``uid_new`` exists and was created with ``PSA_STORAGE_FLAG_WRITE_ONCE``.        
+        * ``uid_new`` exists and ``PSA_STORAGE_FLAG_REPLACE`` is not specified
```
athoelke, line 749: This condition is reported with `PSA_ERROR_ALREADY_EXISTS`

doc/storage/api/api.rst
```diff
@@ -63,7 +63,19 @@ These definitions must be defined in the header file :file:`psa/storage_common.h
     .. field:: psa_storage_create_flags_t flags
         The flags set when the ``uid`` was create
 
+.. struct:: psa_storage_iterator_t
 
+    .. summary::
+        An implantation defined opaque structure containing the context for an iterator.
```
athoelke, line 69:
```suggestion
        An implementation-defined opaque structure containing the context for an iterator. 
```

doc/storage/api/api.rst
```diff
@@ -63,7 +63,19 @@ These definitions must be defined in the header file :file:`psa/storage_common.h
...
 
+    .. summary::
+        An implantation defined opaque structure containing the context for an iterator. 
+        The structure MUST  contain all all the state required by the iterator. 
+        No further state is retained by the implementation. 
+        
+        The structure is initilaised by the `ps_iterator_start` function. 
+        It is modified by the `ps_iterator_next` function. 
+        
+        It is the duty of the calling application to free memory when the iterator is no longer required.
```
athoelke, line 76: As there is no hidden allocation, using the term 'free' might be confusing. Might it be better to state that the application can discard or reuse the iterator object once it has finished using it. This can be before, or after, the iterator has reached the end of the iteration.

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+
+
+.. function:: psa_ps_iterator_start
+
+   .. summary:: 
+       Initialises an iterator that can be used to return a list of uids in the Protected Storage .
+
+   .. param::  psa_storage_iterator_t* context
+       A pointer to a context for this iterator. This is set to a new value on success and is undefined on error. The content of the iterator is implementation defined.  The pointer MUST be valid and MUST point to a block of memory of the size given by the ``PSA_STORAGE_ITERATOR_CTX__SIZE`` macro to which the caller has write access.
```
athoelke, line 773: The final sentence is now redundant

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+       A status indicating the success or failure of the operation.
+
+   .. retval:: PSA_SUCCESS
+       The operation completed successfully.
+   .. retval:: PSA_ERROR_DOES_NOT_EXIST
+       No ``uid`` matches this iteration.
+
+   .. retval:: PSA_ERROR_STORAGE_FAILURE
+       The operation failed because the physical storage has failed (Fatal error).
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
```
athoelke, line 792: No longer necessary

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   Changes to storage by other users MUST NOT affect any open iterations. 
+   
+   ### todo: We need to state what effects are required/denied/permitted (i.e. imp-def) when the caller modifies the storage while iterating. ###
+
+   A caller may initialize multiple iteration contexts at the same time. Each iteration shall be independent. Calling ``psa_ps_iterator_next()`` on one iterator MUST not effect any other open iteration. 
+
+
+.. function:: psa_ps_iterator_next
+   .. summary::
```
athoelke, line 812: Best to have a blank line between the `.. function::` directive and the content

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   Returns the next ``uid`` in this iteration.
```
athoelke, line 814: Requires indenting to make it part of the `.. summary::` directive

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+.. function:: psa_ps_iterator_next
+   .. summary::
+
+   Returns the next ``uid`` in this iteration. 
+
+   .. param::  psa_storage_iterator_t* context
+       A pointer to a context for this iterator as returned by ``psa_ps_iterator_start()`` or updated by a previous call to ``psa_ps_iterator_next``.  The content of the iterator will change on success and is undefined on error.
 
+   .. param:: *psa_storage_uid_t result
```
athoelke, line 819:
```suggestion
   .. param:: psa_storage_uid_t *result
```

doc/storage/api/api.rst
```diff
@@ -649,17 +691,150 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+       Initialises an iterator that can be used to return a list of uids in the Protected Storage .
+
+   .. param::  psa_storage_iterator_t* context
+       A pointer to a context for this iterator. This is set to a new value on success and is undefined on error. The content of the iterator is implementation defined.  The pointer MUST be valid and MUST point to a block of memory of the size given by the ``PSA_STORAGE_ITERATOR_CTX__SIZE`` macro to which the caller has write access. 
+   .. param:: psa_storage_uid_t mask
+       A value used to filter the results included in this iteration. 
+   .. param:: psa_storage_uid_t filter
+       A value used to filter the results included in this iteration. 
+       
+   .. param:: *psa_storage_uid_t result
```
athoelke, line 779:
```suggestion
   .. param:: psa_storage_uid_t *result
```

doc/storage/api/api.rst
```diff
@@ -9,7 +9,7 @@ Status codes
 
 The |API| uses the status code definitions that are shared with the other PSA Certified APIs.
 
-The following elements are defined in :file:`psa/error.h` from :cite-title:`PSA-STAT` (previously defined in :cite:`PSA-FFM`):
+The following elements are defined in :file:`psa/error.h` from :cite-title:`PSA-STAT` (previously defined in :cite:`PSA-FF-M`):
```
athoelke, line 12: This is another artifact of the initial file content not being fully synchronised with upstream.

doc/storage/api/api.rst
```diff
@@ -649,24 +691,165 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   .. retval:: PSA_SUCCESS
+       The operation completed successfully.
+       
+   .. retval:: PSA_ERROR_DOES_NOT_EXIST
+       No ``uid`` matches this iteration.
+
+   .. retval:: PSA_ERROR_STORAGE_FAILURE
+       The operation failed because the physical storage has failed (Fatal error).
+
+   The iterator returns those values where the ``filter_length`` bits of the `uid` matches the bits in ``filter``. If the value of ``filter_length`` is positive, it matches left most bits, if it i nagative, it matches the right most bits.
```
athoelke, line 791: In practice, an implementation is only going to be designed to optimise one of the 'ends' of the uid value. E.g. if the data index is in order of uid value, then optimised filtering is only possible when matching on the most significant bits.

Might it be better to only permit filtering at one end of the uid, so that application usage of the uid values and filtering will always match any optimisation that an implementation can provide?

doc/storage/api/api.rst
```diff
@@ -649,24 +691,165 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   Changes to storage by other users MUST NOT affect any open iterations. 
+
+   A caller may initialize multiple iteration contexts at the same time. Each iteration shall be independent. Calling ``psa_ps_iterator_next()`` on one iterator MUST not effect any other open iteration. 
+   
+   A caller may read the contents of any `uid` with ``psa_ps_get()`` or write with ``psa_ps_set()`` or ``psa_ps_set_extended()`` without invalidating the iteration context. 
+
+   A caller may create a `uid` with ``psa_ps_set()`` or ``psa_ps_create()`` without invalidating the iteration context, provided the `uid` does *NOT* match the filter. However, if the `uid` matches the filter then any later call to `psa_ps_iterator_next()` fails with `PSA_ERROR_DATA_CORRUPT`.
```
athoelke, line 803: These constraints may be challenging to implement. Does this requirement place an arbitrary cost on the implementation?

For example, if an application starts an iteration, and then creates 20 new data items, how does the implementation determine if it must report `PSA_ERROR_DATA_CORRUPT`? It does not have information about the iterator (all state in iterator object in caller), and storing all of the UIDs modified since last iteration, for checking on next ietration, leads to arbitrary volatile storage requirements.

doc/storage/api/api.rst
```diff
@@ -649,24 +691,165 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   
+   A caller may delete a `uid` with ``psa_ps_remove(uid)`` without invalidating the iteration context, provided the `uid` does *NOT* match the filter. However, if the `uid` matches the filter then any later call to `psa_ps_iterator_next()` fails with `PSA_ERROR_DATA_CORRUPT`.
```
athoelke, line 805: With this definition, an application that wants to remove all items that match the iterator, cannot do:

```
if (psa_ps_iterator_start(&iter, filter, bits, &uid)==0) do {
    psa_ps_remove(uid);
} while (psa_pw_iterator_next(&iter, &uid)==0);
```

But instead must use a less intuitive:

```
while (psa_ps_iterator_start(&iter, filter, bits, &uid)==0) {
    psa_ps_remove(uid);
}
```

More troubling are use cases where the removal is dependent on some other test, resulting in non-linear operation as the iterator has to be restarted for each item deleted.

This also complicates use cases where the matching set of items must be modified in a particular way. This can only be done by making two passes - in the first, all UIDs matching the filter are stored in application memory, and then each one can be modified.

doc/storage/api/api.rst
```diff
@@ -649,24 +691,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   Changes to storage by other users MUST NOT affect any open iterations. 
+
+   A caller may initialize multiple iteration contexts at the same time. Each iteration shall be independent. Calling ``psa_ps_iterator_next()`` on one iterator MUST not effect any other open iteration. 
+   
+   An iterator MUST return all data objects whose `uid` matches the filter that are extant when the filter was created, unless these are deleted or renamed before the iteration would return them, or the caller stops before all matching objects have been returned. 
+   
+   A caller may delete a `uid` with ``psa_ps_remove(uid)`` without invalidating the iteration context. the iterator MUST never return a `uid` that has been deleted. However, if the caller is multi-threaded it ia possible another thread may delete a `uid`.
```
athoelke, line 803: That last sentence probably isn't required - except perhaps as a note. It does not constrain the implementation, but is just a reminder to multi-threaded app developers about the risk of race conditions.

athoelke, line 803: * For inline code that includes API identifiers, using the `:code:` role enables the identifiers to be hyperlinked.
* For inline code that is _not_ an API element, use double back-quotes.
* For inline code that is _only_ an API element, use single back-quotes.

If in doubt, use the `:code:` role which will hyperlink if possible, otherwise render as code.

```suggestion
   A caller may delete a ``uid`` with :code:`psa_ps_remove(uid)` without invalidating the iteration context. the iterator MUST never return a ``uid`` that has been deleted. However, if the caller is multi-threaded it ia possible another thread may delete a ``uid``. 
```

doc/storage/api/api.rst
```diff
@@ -649,24 +691,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   Changes to storage by other users MUST NOT affect any open iterations. 
+
+   A caller may initialize multiple iteration contexts at the same time. Each iteration shall be independent. Calling ``psa_ps_iterator_next()`` on one iterator MUST not effect any other open iteration. 
+   
+   An iterator MUST return all data objects whose `uid` matches the filter that are extant when the filter was created, unless these are deleted or renamed before the iteration would return them, or the caller stops before all matching objects have been returned. 
+   
+   A caller may delete a `uid` with ``psa_ps_remove(uid)`` without invalidating the iteration context. the iterator MUST never return a `uid` that has been deleted. However, if the caller is multi-threaded it ia possible another thread may delete a `uid`. 
+   
+   A caller may read the contents of any `uid` with ``psa_ps_get()`` or write with ``psa_ps_set()`` or ``psa_ps_set_extended()`` without invalidating the iteration context.
```
athoelke, line 805:
```suggestion
   A caller may read the contents of any ``uid`` with `psa_ps_get()` or write with `psa_ps_set()` or `psa_ps_set_extended()` without invalidating the iteration context. 
```

etc... (there are other instances)

doc/storage/api/api.rst
```diff
@@ -649,24 +691,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   
+   An iterator MUST return all data objects whose `uid` matches the filter that are extant when the filter was created, unless these are deleted or renamed before the iteration would return them, or the caller stops before all matching objects have been returned. 
+   
+   A caller may delete a `uid` with ``psa_ps_remove(uid)`` without invalidating the iteration context. the iterator MUST never return a `uid` that has been deleted. However, if the caller is multi-threaded it ia possible another thread may delete a `uid`. 
+   
+   A caller may read the contents of any `uid` with ``psa_ps_get()`` or write with ``psa_ps_set()`` or ``psa_ps_set_extended()`` without invalidating the iteration context. 
+
+   A caller may create a `uid` with ``psa_ps_set()`` or ``psa_ps_create()`` without invalidating the iteration context. However, the iterator is NOT guaranteed to return the new object, `uid`, the behaviour is dependent on both implementation and identity. In particular, the iterator is not expected to return `uid` if the iteration is already past the point at which it would naturally be returned.
```
athoelke, line 807: I think 'For example' works better than 'In particular' - as this is not normative text.

doc/storage/api/api.rst
```diff
@@ -649,24 +691,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   
+   A caller may read the contents of any `uid` with ``psa_ps_get()`` or write with ``psa_ps_set()`` or ``psa_ps_set_extended()`` without invalidating the iteration context. 
+
+   A caller may create a `uid` with ``psa_ps_set()`` or ``psa_ps_create()`` without invalidating the iteration context. However, the iterator is NOT guaranteed to return the new object, `uid`, the behaviour is dependent on both implementation and identity. In particular, the iterator is not expected to return `uid` if the iteration is already past the point at which it would naturally be returned. 
+   
+   A caller may call `psa_ps_rename(uid, uid_new)` without invalidating the iteration context. The iterator must not return `uid`. The iterator is not guaranteed to return `uid_new`, the behaviour is dependent on both implementation and identity.  
+
+   The following code snippet uses a linked list to store the matching files before iterating over that list and removing them.
```
athoelke, line 811: This comment is out of date as the linked list has been removed.

athoelke (changes requested):
Largely there I think in terms of the API design.

I think we might need to decide how to deal with the optionality of the new APIs. (and maybe review what is said in the spec about `psa_ps_create()` and `psa_ps_set_extended()`). This includes whether an implementation that claims to be v1.1 (presumably what this update will be), is permitted to not support the iterator API. If so, how should we articulate this in the API specification, consistently across all affected APIs.

The example needs to be improved to be recognisably, and mostly compilable, C code:
* Use the `xref` format for the code block (results in hyperlinked API elements)
* Data types
* Bracketing
* Extracting checking the return code from `psa_ps_iterator_next()`

US English spelling please.

athoelke:
The merge of main back into this branch has no effect because Git believes that the relevant changes have previously been merged, and this branch has then made changes after that merge.

To repair the inadvertent 'revert' changes to some of the source files will require more effort.

* You could add a new commit to change them back to the current content on main. That isn't ideal in terms of history, but it works.
* Otherwise you need to find the commits in the current branch that introduce those changes, and then edit those commits to no longer do so. An interactive rebase can make that process a bit easier.
* Or start with the first option, then use an interactive rebase to move the 'fix' commit immediately after the commit that makes the incorrect change, and mark it as 'fixup' in the interactive rebase file.

MarcusJGStreets:
> Added in PS storage initially. If thee is need for them in ITS, the same additions can be made there

Now added to ITS

doc/storage/api/api.rst
athoelke: This has been broken again. Related to the previous commit about build errors which also undoes changes made in main in January,

doc/storage/api/api.rst
```diff
@@ -313,6 +356,135 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
...
+   .. summary::
+       Initializes an iterator that can be used to return a list of ``uid`` values in the Internal Trusted Storage.
+       
+       This function must be fully defined if `PSA_STORAGE_SUPPORT_ITERATION` is true.
+       
+       If `PSA_STORAGE_SUPPORT_ITERATION` is false, then this function SHALL always return ``PSA_ERROR_NOT_SUPPORTED``
+
+
+   .. param::  psa_its_storage_iterator_t* context
+       A pointer to a context for this iterator. The pointer may be NULL. This is set to a new value on success and is undefined on error. The content of the iterator is implementation defined.
```
athoelke, line 370: The optional `NULL` value looks odd, as it would seem that the entire purpose of starting an iteration would be to iterate.

There is no mention of how the function behaves if no iterator object is provided, or why a caller would ever want to do this. If there is a use case (I can think of a couple) where this is a better approach than provided an iterator object, it would benefit from an explanation and/or code example in the description.

athoelke: The same applies to `psa_ps_iterator_start()`.

athoelke (changes requested):
I have some more changes to work through... but here are an initial set to consider.

MarcusJGStreets:
Updates with clarifications that iterator start returns a pointer to a new context. 
And split the support iterator macro into ITS and PS for compile time detection.

doc/storage/api/api.rst
```diff
@@ -658,15 +842,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
 
+    * Data in the ranges 0 to ``data_offset`` is not modified.
 
+    * If ``data_offset + data_length < size`` then data in the range ``data_offset + data_length` to `size`` is not modified.
 
-.. function:: psa_ps_get_support
 
-    .. return:: uint32_t
+
+.. function:: psa_ps_rename
```
athoelke, line 851: Is there a reason why rename is not an API that we want to permit on the ITS storage?

doc/storage/api/api.rst
```diff
@@ -313,6 +339,139 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
...
+   .. retval:: PSA_SUCCESS
+       The operation completed successfully.
+
+   .. retval:: PSA_ERROR_DOES_NOT_EXIST
+       The iterator has returned all the uids that match this iteration.
+
+   .. retval:: PSA_ERROR_STORAGE_FAILURE
+       The operation failed because the physical storage has failed (Fatal error).
+
+   .. retval:: PSA_ERROR_DATA_CORRUPT
```
athoelke, line 408: This contradicts the description text relating to the stability of the iterator when the underlying data storage is modified by the application between calls?

doc/storage/api/api.rst
```diff
@@ -350,10 +509,29 @@ These definitions must be defined in the header file :file:`psa/protected_storag
 
     It will be incremented in small updates that are unlikely to include breaking changes.
 
+.. macro:: PSA_PS_SUPPORT_ITERATION
+    (1u << 2)
+
+    Flag indicating that `psa_ps_iterator_start` and `psa_ps_iterator_next` are supported.
+
+.. struct:: psa_ps_storage_iterator_t
```
athoelke, line 517: Make this a typedef, as per the ITS iterator:

```suggestion
.. struct:: psa_ps_storage_iterator_t
    :type:
```

doc/storage/api/api.rst
```diff
@@ -350,10 +509,29 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+    .. summary::
+        An implementation-defined opaque structure containing the context for an iterator.
+        The structure MUST contain all all the state required by the iterator.
+        That is, further state MUST NOT be retained by the implementation.
+
+        The structure is initilaised by the ``ps_iterator_start()`` function.
+        It is modified by the ``ps_iterator_next()`` function.
+
+        the caller can discard or reuse the iterator object once it has finished using it. This can be before, or after, the iterator has reached the end of the iteration.
```
athoelke, line 527:
```suggestion
        The caller can discard or reuse the iterator object once it has finished using it. This can be before, or after, the iterator has reached the end of the iteration.
```

doc/storage/api/api.rst
```diff
@@ -350,10 +509,29 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+        The header file is only required to define this structure if PSA_STORAGE_SUPPORT_ITERATION is true.
```
athoelke, line 529: Discussion on optionality as per ITS iterator.

doc/storage/api/api.rst
```diff
@@ -658,15 +842,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+.. function:: psa_ps_iterator_start
+
+   .. summary::
+       Initializes an iterator that can be used to return a list of uids in the Protected Storage.
+       This function must be fully defined if iteration is supported.
+       If iteration is not supported, then this function SHALL be defined in the header - potentially as an inline function - but SHALL return ``PSA_ERROR_NOT_SUPPORTED``
+
+
+   .. param::  psa_ps_storage_iterator_t *context
+       The location in which the function should store a pointer to the newly created context for this iterator.
```
athoelke, line 916:
```suggestion
       The location in which the function stores the iteration context.
```

doc/storage/api/api.rst
```diff
@@ -658,15 +842,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   A caller may read the contents of any ``uid`` with `psa_ps_get()` or write with `psa_ps_set()` or `psa_ps_set_extended()` without invalidating the iteration context.
+
+   A caller may create a ``uid`` with `psa_ps_set()` or `psa_ps_create()` without invalidating the iteration context. However, the iterator is NOT guaranteed to return the new object, ``uid``, the behavior is dependent on both implementation and identity. In particular, the iterator is not expected to return ``uid`` if the iteration is already past the point at which it would naturally be returned.
+
+   A caller may call ``psa_ps_rename(uid, uid_new)`` without invalidating the iteration context. The iterator must not return ``uid``. The iterator is not guaranteed to return ``uid_new``, the behavior is dependent on both implementation and identity.
+
+   The following code snippet uses a linked list to store the matching files before iterating over that list and removing them.
```
athoelke, line 961: Out-dated comment

doc/storage/api/api.rst
```diff
@@ -313,6 +339,139 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
...
+       The operation failed because the contents of the iteration have changed. That is a ``uid`` matching the filter has either been created or deleted.
+
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+       The operation failed because either:
+
+       * The provided context is not valid.
+
+       * The caller cannot access the memory at ``result``
+
+   The iterator returns those values where the ``filter_length`` bits of the ``uid`` matches the left most bits in ``filter``.
```
athoelke, line 418: This and the following descriptive text would be better placed in the `psa_its_iterator_start()` function. This function can refer to the start function for description of the behavior of the iterator if the storage is modified between iterator calls.

doc/storage/api/api.rst
```diff
@@ -658,15 +842,193 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   .. retval:: PSA_ERROR_DATA_CORRUPT
+       The operation failed because the contents of the iteration have changed. That is a ``uid`` matching the filter has either been created or deleted.
+
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+       The operation failed because either:
+
+       * The provided context is not valid.
+
+       * The caller cannot access the memory at ``result``
+
```
athoelke, line 1021: Probably worth at least a single line in the function description to refer to the start function description of how the iterator works and how to use it.

athoelke (changes requested):
A number of minor nits, and a handful of more significant issues to resolve:

* The **about.txt** file is still broken (reverts changes in `main`)
* How do we want to handle optionality in the API (see specific comments for issue & options)
* Iterator description for ITS should be in the start function, not next.
* Structures should also be type[def]s. This is easy to add - and we should probably do the same for the legacy `psa_storage_info_t` structure (this will be backwards compatible).
* The usage example code is not quite valid C, and the introductory text is out of date.

doc/storage/api/api.rst
```diff
@@ -305,14 +332,195 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
     .. retval:: PSA_ERROR_STORAGE_FAILURE
         The operation failed because the physical storage has failed (Fatal error).
 
-    Deletes the data from internal storage.
+   Introduced in Version 1.1.
```
athoelke, line 335: `psa_its_remove()` has been in the API since 1.0. Lines 335-337 are not required here.

doc/storage/api/api.rst
```diff
@@ -305,14 +332,195 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
...
+
+   .. retval:: PSA_ERROR_GENERIC_ERROR
+        The operation failed because of an unspecified internal failure.
+
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The operation failed because either:
+
+        * ``uid`` is ``0``.
+        * ``uid_new`` is ``0``
+        * the ``psa_storage_rename_flags_t`` has a value set other than `PSA_STORAGE_FLAG_REPLACE`
```
athoelke, line 382: This should be about the `rename_flags` parameter, not the type.

```suggestion
        * ``rename_flags`` has a value set other than `PSA_STORAGE_FLAG_REPLACE`.
```

doc/storage/api/api.rst
```diff
@@ -305,14 +332,195 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
...
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+       The operation failed because either:
+
+       * The provided context is not valid.
+
+       * The caller cannot access the memory at ``result``
+
+   Introduced in Version 1.1. 
+   
+   This function takes an iterator that was returned by the `psa_its_iterator_start` function and searches the storage for the next ``uid`` that matches the filters defined in that function call. If a ``uid`` matching the filter exists, the function updates the ``result`` and ``context`` parameters and retursn PSA_SUCCESS. If not it reruns an error.
```
athoelke, line 505: Spelling: 'reruns' -> 'returns'

doc/storage/api/api.rst
```diff
@@ -658,15 +890,170 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+       The operation failed because either:
+
+       * The provided context is not valid.
+
+       * The caller cannot access the memory at ``result``
+
+   Introduced in Version 1.1. 
+   
+   This function takes an iterator that was returned by the `psa_ps_iterator_start` function and searches the storage for the next ``uid`` that matches the filters defined in that function call. If a ``uid`` matching the filter exists, the function updates the ``result`` and ``context`` parameters and retursn PSA_SUCCESS. If not it reruns an error.
```
athoelke, line 1046: Spelling: 'reruns' -> 'returns'

doc/storage/api/api.rst
```diff
@@ -658,15 +890,170 @@ These definitions must be defined in the header file :file:`psa/protected_storag
...
+
+   .. retval:: PSA_ERROR_GENERIC_ERROR
+        The operation failed because of an unspecified internal failure.
+
+   .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The operation failed because either:
+
+        * ``uid`` is ``0``.
+        * ``uid_new`` is ``0``
+        * the ``psa_storage_rename_flags_t`` has a value set other than `PSA_STORAGE_FLAG_REPLACE`
```
athoelke, line 932:
```suggestion
        * ``rename_flags`` has a value set other than `PSA_STORAGE_FLAG_REPLACE`.
```

doc/storage/api/api.rst
```diff
@@ -140,6 +148,24 @@ These definitions must be defined in the header file :file:`psa/internal_trusted
 
     It will be incremented in small updates that are unlikely to include breaking changes.
 
+.. macro:: PSA_ITS_SUPPORT_ITERATION
+    (1u << 2)
+
+    Flag indicating that `psa_its_iterator_start()` and  `psa_its_iterator_next()` are supported.
+
+.. struct:: psa_its_storage_iterator_t
```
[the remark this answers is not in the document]
athoelke: I've changed my mind: as this is a fully implementation-defined type, we should probably use the same pattern as the Crypto API for such types, and just use a typedef in this specification. For example, see [the definition of `psa_key_attributes_t`](https://arm-software.github.io/psa-api/crypto/1.2/api/keys/attributes.html#c.psa_key_attributes_t). This gives the implementation full flexibility in the definition of the type.

To do this, replace:

```
.. struct:: psa_its_storage_iterator_t
    :type:
```

with

```
.. typedef:: /* implementation-defined type */ psa_its_storage_iterator_t
```

Likewise for the PS iterator type.

In the Crypto API, we then refer to these as 'objects' rather than 'structures' - i.e. use 'iterator object' in this spec when describing a variable of this type.

athoelke (changes requested):
Only some minor issues remain

athoelke:
While working through the issues for this addition: I wonder if we should improve the text for the optional 1.0 functions `psa_ps_create()` and `psa_ps_set_extended()`?

The current wording is not clear whether an implementation that does not support these functions must provide definitions of them in a header file (perhaps with inline implementations), and so ensure that an application that references these functions will compile?

What does TF-M do?

PR #156, opened 2024-01-19
Add export format for SPAKE2+ key pairs

### LINKED ISSUE #153: Define the export format for a SPAKE2+ key pair
_From Oberon (Referencing the [SPAKE2+ draft2](https://github.com/ARM-software/psa-api/files/13530157/AES0058-PSA_Certified_Crypto_API-1.2_PAKE_Extension-bet.2-draft2-spake2plus.pdf) rendering of the PAKE Extension):_

The spec draft states on page 54, section 2.3.4 SPAKE2+ keys:

> The calculation of w0, w1, and L then proceeds as described in the RFC.
>
> **Implementation note**
>
>    The values of w0 and w1 are required as part of the SPAKE2+ key pair.
>    It is IMPLEMENTATION DEFINED whether L is computed during key derivation, and stored as part of
>    the key pair; or only computed when required from the key pair.

However, a defined export format for secret keys is required to support:
* key import and export,
* standard key data handling in the API of transparent drivers.

Even if import and export of secret PAKE keys is not needed in typical use cases, it should be defined for completeness, for testing, and to support predefined keys.

athoelke:
Fixes #153

### DIFF
--- a/doc/ext-pake/api/pake.rst
+++ b/doc/ext-pake/api/pake.rst
@@ -1752,7 +1752,7 @@ SPAKE2+ keys
     .. param:: curve
         A value of type :code:`psa_ecc_family_t` that identifies the Elliptic curve family to be used.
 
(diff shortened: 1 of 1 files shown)


athoelke:
I've added a warning as described in the comment to #153

PR #159, opened 2024-01-30
Clarification and relaxation of the flow for PAKE protocols

### LINKED ISSUE #157: Ordering requirements/flexibility in PAKE operations
At the moment the PAKE operations for J-PAKE and SPAKE2+ present an ordering of the calls to the API to perform the PAKE protocol. However, the specification is not clear about whether these are strict requirements relating to the call sequence (which restricts the application behavior), or if there is some flexibility.

There seems to be a reasonable case for documenting where an implementation must provide some flexibility in the ordering, to satisfy application use cases; as well as making it clear where the ordering is essential to preserve the correct operation and protocol behavior.

### J-PAKE

As an example of where flexibility would be valuable for applications, consider J-PAKE being conducted between participants _A_ and _B_, trying to minimize the number of messages. (The following description refers to steps as numbered in p40-42 of the [Beta.2 Release candidate PDF](https://github.com/ARM-software/psa-api/files/13890452/AES0058-PSA_Certified_Crypto_API-1.2_PAKE_Extension-bet.2-rc1.pdf) in #141)

* _A_ sets up a PAKE operation (step 1 in the specification) and extracts the six round-1 output values (step 2).
* _A_ then sends these values to _B_.
* When _B_ receives the message from _A_, it sets up its own PAKE operation (step 1). **What happens next:**
  - **MUST** _B_ extract the six round-1 outputs (step 2), and only then provide the six round-1 inputs received from _A_ (step 3)? **- OR**
  - Is _B_ **permitted** (instead) to first provide the six round-1 inputs received from _A_ (step 3), and then extract the six round-1 outputs (step 2)?
* _B_ now also extracts the six round-2 outputs (step 4).
* _B_ sends all of its round-1 and round-2 outputs to _A_.
* When _A_ receives the message from _B_, it MUST input the round-1 values (step 3) before extracting the six round-2 values (step 4), as the latter are [computationally] dependent on the round-1 inputs. **However:**
  - **MUST** _A_ extract the six round-2 outputs (step 4), and only then provide the six round-2 inputs received from _B_ (step 5)? **- OR**
  - Is _A_ **permitted** (instead) to first provide the six round-2 inputs received from _B_ (step 5), and then extract the six round-2 outputs (step 4)?
* _A_ sends its round-2 outputs to _B_.
* When _B_ receives the message from _A_, it inputs the round-2 values (step 5).
* Both _A_ and _B_ can now extract the shared key (step 6)

From an application point of view, it is helpful to be able to input data received in a message before extracting data to send in a reply - even when the outputs do not depend on the inputs. I.e. Application developers would benefit from the flexibility to carry out step 3 before step 2, and/or step 5 before step 4; note that the API depends on inputs being provided in-order (step 2 before step 4), and outputs extracted in-order (step 3 before step 5), and the protocol requires step 3 before step 4, and steps 4 and 5 before step 6.

### SPAKE2+

There is a similar issue for SPAKE2+. In this case the flow in the specification (p48-51) shows the round-1 _shareP_ from the Prover/Client being input the Verifier/Server operation before extracting the Verifier round-1 _shareV_. Is this ordering mandated, or can the Verifier application extract _shareV_ before input of _shareP_?

There is another related issue for SPAKE2+, although in this case, the argument is about which participant might need more protection against online dictionary attacks. The operational flow (in p48-51 of the same document) matches the SPAKE2+ RFC, where it shows a 3-message protocol, where _shareV_ and _confirmV_ are transmitted together by the Verifier/Server. It can be imagined that for some uses of the SPAKE2+ algorithm, it is more important to protect the Verifier/Server from an online dictionary attack, and have the Prover/Client send it's _confirmP_ value to the Verifier before the Verifier sends _confirmV_. In that case, the Prover operation must be output _confirmP_, before _confirmV_ is input (in contrast to the order shown in the specification).

athoelke:
Permit flexibility in the flow for J-PAKE:
* Allow the application to either input-before-output or output-before-input in each round of the protocol.

Correct the SPAKE2+ flow description:
* Flow diagram had reversed two of the calls in the Prover
* Clarify that the presented sequence of calls is mandatory
* Fix some incorrect comments in the demonstration code

Fixes #157


athoelke:
@silabs-Kusumit, @silabs-hannes, @oberon-sk, @bheeb, @yanesca: please review these [hopefully] final changes for the PAKE API before we publish the updated specification.

bheeb:
Looks good to me

silabs-Kusumit:
Looks good to me

yanesca:
Looks good to me

PR #175, opened 2024-02-22
Fix Jake sample code for setting key attributes

mswarowsky:
The function to set the key attributes is psa_set_key_xxx not psa_key_set_xxx so changed it to the correct version


athoelke (approved):
Thank you! - and LGTM

PR #178, opened 2024-03-01
Add definitions required by SRP-6.

oberon-sk:
The _PSA Certified Crypto API v1.2 PAKE Extension_ requires some additional definitions to support SRP-6 as defined by RFC 2945 and RFC 5054 to be embedded in, or included by, _psa/crypto.h_.

This PR provides a proposal.

Signed-off-by: Oberon microsystems

### DIFF
--- a/headers/crypto/1.2/psa/crypto-srp.h
+++ b/headers/crypto/1.2/psa/crypto-srp.h
@@ -1,0 +1,182 @@
+/* This file contains reference definitions for implementations of 
+ * SRP-6 in addition to the PSA Certified Crypto API v1.2 PAKE Extension
+ *
+ * These definitions must be embedded in, or included by, psa/crypto.h
+ */
+
+#define PSA_KEY_TYPE_SRP_KEY_PAIR_BASE          ((psa_key_type_t) 0x7700)
+#define PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE        ((psa_key_type_t) 0x4700)
+#define PSA_KEY_TYPE_SRP_GROUP_MASK             ((psa_key_type_t) 0x00ff)
+
+ /** SRP key pair. Both the client and server key.
+ *
+ * The size of a SRP key is the size associated with the Diffie-Hellman
+ * group. See the documentation of each Diffie-Hellman group for details.
+ * To construct a SRP key pair, the password hash must be imported.
+ * The corresponding public key (password verifier) can be exported using
+ * psa_export_public_key(). See also #PSA_KEY_TYPE_SRP_PUBLIC_KEY().
+ *
+ * \param group A value of type ::psa_dh_family_t that identifies the
+ *              Diffie-Hellman group to be used.
+ */
+#define PSA_KEY_TYPE_SRP_KEY_PAIR(group) \
+    ((psa_key_type_t) (PSA_KEY_TYPE_SRP_KEY_PAIR_BASE | (group)))
+
+ /** SRP public key. The server key (password verifier).
+ *
+ * The size of an SRP public key is the same as the corresponding private
+ * key. See #PSA_KEY_TYPE_SRP_KEY_PAIR() and the documentation of each
+ * Diffie-Hellman group for details.
+ * To construct a SRP public key, it must be imported. The key size
+ * in attributes must not be zero.
+ *
+ * \param group A value of type ::psa_dh_family_t that identifies the
+ *              Diffie-Hellman group to be used.
+ */
+#define PSA_KEY_TYPE_SRP_PUBLIC_KEY(group) \
+    ((psa_key_type_t) (PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE | (group)))
+
+ /** Whether a key type is a SRP key (pair or public-only). */
+#define PSA_KEY_TYPE_IS_SRP(type)                                 \
+    ((PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(type) &                 \
+      ~PSA_KEY_TYPE_SRP_GROUP_MASK) ==                            \
+      PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE)
+ /** Whether a key type is a SRP key pair. */
+#define PSA_KEY_TYPE_IS_SRP_KEY_PAIR(type)                        \
+    (((type) & ~PSA_KEY_TYPE_SRP_GROUP_MASK) ==                   \
+     PSA_KEY_TYPE_SRP_KEY_PAIR_BASE)
+ /** Whether a key type is a SRP public key. */
+#define PSA_KEY_TYPE_IS_SRP_PUBLIC_KEY(type)                      \
+    (((type) & ~PSA_KEY_TYPE_SRP_GROUP_MASK) ==                   \
+     PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE)
+ /** Extract the curve from a SRP key type. */
+#define PSA_KEY_TYPE_SRP_GET_FAMILY(type)                         \
+    ((psa_ecc_family_t) (PSA_KEY_TYPE_IS_SRP(type) ?              \
+                         ((type) & PSA_KEY_TYPE_SRP_GROUP_MASK) : \
+                         0))
+
+/** The Secure Remote Passwort key exchange (SRP) algorithm.
+ *
+ * This is SRP-6 as defined by RFC 2945 and RFC 5054, instantiated with the
+ * following parameters:
+ *
+ * - The group is defined over a finite field using a secure prime.
+ * - A cryptographic hash function.
+ *
+ * To select these parameters and set up the cipher suite, call these functions:
+ *
+ * \code
+ * psa_pake_cipher_suite_t cipher_suite = PSA_PAKE_CIPHER_SUITE_INIT;
+ * psa_pake_cs_set_algorithm(cipher_suite, PSA_ALG_SRP_6(hash));
+ * psa_pake_cs_set_primitive(&cipher_suite,
+ *                           PSA_PAKE_PRIMITIVE(type, family, bits));
+ * \endcode
+ *
+ * After initializing a SRP operation, call:
+ *
+ * \code
+ * psa_pake_setup(operation, password, cipher_suite);
+ * psa_pake_set_role(operation, ...);
+ * psa_pake_set_user(operation, ...);
+ * \endcode
+ * 
+ * The password provided to the client side must be of type
+ * #PSA_KEY_TYPE_SRP_KEY_PAIR.
+ * The password provided to the server side must be of type
+ * #PSA_KEY_TYPE_SRP_PUBLIC_KEY.
+ *
+ * The role set by \c psa_pake_set_role() must be either
+ * \c PSA_PAKE_ROLE_CLIENT or \c PSA_PAKE_ROLE_SERVER.
+ * 
+ * For the SRP client key exchange call the following functions in any order:
+ * \code
+ * // get salt
+ * psa_pake_input(operation, #PSA_PAKE_STEP_SALT, ...);
+ * // get server key
+ * psa_pake_input(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * // write client key
+ * psa_pake_output(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * \endcode
+ *
+ * For the SRP server key exchange call the following functions in any order:
+ * \code
+ * // get salt
+ * psa_pake_input(operation, #PSA_PAKE_STEP_SALT, ...);
+ * // get client key
+ * psa_pake_input(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * // write server key
+ * psa_pake_output(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * \endcode
+ *
+ * For the client proof phase call the following functions in this order:
+ * \code
+ * // send M1
+ * psa_pake_input(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // receive M2
+ * psa_pake_output(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // Get secret
+ * psa_pake_get_shared_key()
+ * \endcode
+ *
+ * For the server proof phase call the following functions in this order:
+ * \code
+ * // receive M1
+ * psa_pake_output(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // send M2
+ * psa_pake_input(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // Get secret
+ * psa_pake_get_shared_key()
+ * \endcode
+ *
+ * The shared secret that is produced by SRP is pseudorandom. Although
+ * it can be used directly as an encryption key, it is recommended to use
+ * the shared secret as an input to a key derivation operation to produce
+ * additional cryptographic keys.
+ */
+#define PSA_ALG_SRP_6_BASE                      ((psa_algorithm_t) 0x0a000300)
+#define PSA_ALG_SRP_6(hash_alg) (PSA_ALG_SRP_6_BASE | ((hash_alg) & PSA_ALG_HASH_MASK))
+#define PSA_ALG_IS_SRP_6(alg) (((alg) & ~PSA_ALG_HASH_MASK) == PSA_ALG_SRP_6_BASE)
+
+/** The salt.
+ *
+ * The format for both input and output at this step is plain binary data.
+ */
+#define PSA_PAKE_STEP_SALT                      ((psa_pake_step_t)0x05)
+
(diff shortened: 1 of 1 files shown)


headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * SRP-6 in addition to the PSA Certified Crypto API v1.2 PAKE Extension
+ *
+ * These definitions must be embedded in, or included by, psa/crypto.h
+ */
+
+#define PSA_KEY_TYPE_SRP_KEY_PAIR_BASE          ((psa_key_type_t) 0x7700)
+#define PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE        ((psa_key_type_t) 0x4700)
+#define PSA_KEY_TYPE_SRP_GROUP_MASK             ((psa_key_type_t) 0x00ff)
+
+ /** SRP key pair. Both the client and server key.
```
athoelke, line 11: The definition of these SRP-6 keys is technically equivalent to FFDH keys, in terms of key content and calculation of the public key from the private key.

However, there may be a good reason for maintaining a separate key type definition for SRP-6 keys instead of reusing the existing `PSA_KEY_TYPE_DH_KEY_PAIR()`: The derivation of an SRP-6 key is different to the default method for FFDH keys in the Crypto API.
* An FFDH key-pair is derived by repeated sampling of the KDF output to find a value in the required range [1, N-1].
* A SRP-6 key-pair is constructed by zero-padding the fixed-size output of the password-hash. The original SRP hash outputs only 160 bits (for SHA1).

Note to self: this may be worth a rationale comment when defined in the specification.

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * The size of a SRP key is the size associated with the Diffie-Hellman
+ * group. See the documentation of each Diffie-Hellman group for details.
+ * To construct a SRP key pair, the password hash must be imported.
+ * The corresponding public key (password verifier) can be exported using
+ * psa_export_public_key(). See also #PSA_KEY_TYPE_SRP_PUBLIC_KEY().
+ *
+ * \param group A value of type ::psa_dh_family_t that identifies the
```
athoelke, line 19: Are there specific DH groups that could/should be used here?

* Crypto API currently just defines the RFC 7919 family
* This header defines another RFC 3526 family
* The SRP-for-TLS reference (RFC 5054) defines another set of FFDH groups

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
+/* This file contains reference definitions for implementations of 
+ * SRP-6 in addition to the PSA Certified Crypto API v1.2 PAKE Extension
+ *
+ * These definitions must be embedded in, or included by, psa/crypto.h
+ */
+
+#define PSA_KEY_TYPE_SRP_KEY_PAIR_BASE          ((psa_key_type_t) 0x7700)
```
athoelke, line 7: This value works well with the current set of DH key values: it has the same parity as the `PSA_KEY_TYPE_DH_KEY_PAIR()` base value, so existing DH family values can be used.

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+     PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE)
+ /** Extract the curve from a SRP key type. */
+#define PSA_KEY_TYPE_SRP_GET_FAMILY(type)                         \
+    ((psa_ecc_family_t) (PSA_KEY_TYPE_IS_SRP(type) ?              \
+                         ((type) & PSA_KEY_TYPE_SRP_GROUP_MASK) : \
+                         0))
+
+/** The Secure Remote Passwort key exchange (SRP) algorithm.
+ *
+ * This is SRP-6 as defined by RFC 2945 and RFC 5054, instantiated with the
```
athoelke, line 60: Is this SRP-6 or SRP-6a? the latter calculates the multiplier value *k* using `Hash(N || PAD(g))`, instead of using the fixed value `3`. This distinction is not always correctly defined in references to this algorithm.

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * - The group is defined over a finite field using a secure prime.
+ * - A cryptographic hash function.
+ *
+ * To select these parameters and set up the cipher suite, call these functions:
+ *
+ * \code
+ * psa_pake_cipher_suite_t cipher_suite = PSA_PAKE_CIPHER_SUITE_INIT;
+ * psa_pake_cs_set_algorithm(cipher_suite, PSA_ALG_SRP_6(hash));
+ * psa_pake_cs_set_primitive(&cipher_suite,
+ *                           PSA_PAKE_PRIMITIVE(type, family, bits));
```
athoelke, line 72: The parameters here could be more fully specified - something to fix in the specification itself.

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * \endcode
+ *
+ * After initializing a SRP operation, call:
+ *
+ * \code
+ * psa_pake_setup(operation, password, cipher_suite);
+ * psa_pake_set_role(operation, ...);
+ * psa_pake_set_user(operation, ...);
```
athoelke, line 80: `psa_pake_set_user()` sets the local participant identity. So for the server role, `psa_pake_set_peer()` would need to be called. However, unless it is used in the key confirmation computation (see other comments), the pake operation does not require this value.

There is no server identity in this PAKE, setting a value should probably return an error?

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * \endcode
+ * 
+ * The password provided to the client side must be of type
+ * #PSA_KEY_TYPE_SRP_KEY_PAIR.
+ * The password provided to the server side must be of type
+ * #PSA_KEY_TYPE_SRP_PUBLIC_KEY.
```
athoelke, line 86: In SPAKE2+ we permitted the use of the key-pair here as well. Should we permit the same here, or was that a mistake?

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * The role set by \c psa_pake_set_role() must be either
+ * \c PSA_PAKE_ROLE_CLIENT or \c PSA_PAKE_ROLE_SERVER.
+ * 
+ * For the SRP client key exchange call the following functions in any order:
```
athoelke, line 91: We were rather more strict about the ordering in the SPAKE2+ definition. Is the flexibility here important? - e.g. to support different protocol flows regarding transmission of User Id and receipt of salt from the server?

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * \code
+ * // get salt
+ * psa_pake_input(operation, #PSA_PAKE_STEP_SALT, ...);
```
athoelke, line 94: The Client only requires the salt for password hashing (done outside the PAKE operation, to prepare the SRP-6 key-pair), and for *some* definitions of the key confirmation values. i.e. this might be unnecessary depending on the definition of the key confirmation values (see other comments).

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * // get server key
+ * psa_pake_input(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * // write client key
+ * psa_pake_output(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * \endcode
+ *
+ * For the SRP server key exchange call the following functions in any order:
+ * \code
+ * // get salt
+ * psa_pake_input(operation, #PSA_PAKE_STEP_SALT, ...);
```
athoelke, line 104: The Server only requires the salt for *some* definitions of the key confirmation values. i.e. this might be unnecessary depending on the definition of the key confirmation values (see other comments).

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * // write server key
+ * psa_pake_output(operation, #PSA_PAKE_STEP_KEY_SHARE, ...);
+ * \endcode
+ *
+ * For the client proof phase call the following functions in this order:
+ * \code
+ * // send M1
+ * psa_pake_input(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // receive M2
+ * psa_pake_output(operation, #PSA_PAKE_STEP_CONFIRM, ...);
```
athoelke, line 116: The definition of M2 needs to be clarified. I can find more than 3 different formulations across different SRP-6 documentation.

The definition of M1 and M2 may result in not requiring some of the inputs to the operation (client identity and salt).

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * // Get secret
+ * psa_pake_get_shared_key()
+ * \endcode
+ *
+ * For the server proof phase call the following functions in this order:
+ * \code
+ * // receive M1
+ * psa_pake_output(operation, #PSA_PAKE_STEP_CONFIRM, ...);
```
athoelke, line 124: The definition of M1 needs to be clarified. I can find different formulations across different SRP-6 documentation.

The definition of M1 and M2 may result in not requiring some of the inputs to the operation (client identity and salt).

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * \endcode
+ *
+ * For the client proof phase call the following functions in this order:
+ * \code
+ * // send M1
+ * psa_pake_input(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // receive M2
+ * psa_pake_output(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // Get secret
+ * psa_pake_get_shared_key()
```
athoelke, line 118: The formulation of the shared key needs to be clarified.

* In RFC 2945, it uses an interleaved SHA1 to construct an unbiased pseudo-random 320-bit value from the pre-master secret (biased output of a function on the DH group).
* In RFC 5054, the description stops at the pre-mater secret (S), a biased output of a function on the DH group.
* Elsewhere it just uses Hash(S), for some value of 'Hash'.

Any of these might be appropriate...

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+ * It calculates the password hash h as
+ * h = H(salt || H(u || ":" || p)
+ * where H is the given hash algorithm.
+ * 
+ * This key derivation algorithm uses the following inputs, which must be
+ * provided in the following order:
+ * - #PSA_KEY_DERIVATION_INPUT_INFO is the user id.
+ * - #PSA_KEY_DERIVATION_INPUT_PASSWORD is the password.
+ * - #PSA_KEY_DERIVATION_INPUT_SALT is the salt.
+ * The output has to be read as a key of type PSA_KEY_TYPE_SRP_KEY_PAIR.
```
athoelke, line 166: I presume this means that the intended use is to call `psa_key_derivation_output_key()` requesting a key of that type, with a key size that matches the bit-size of the DH group used for the primitive in the PAKE cipher-suite?

headers/crypto/1.2/psa/crypto-srp.h
```diff
@@ -0,0 +1,182 @@
...
+
+#define PSA_KEY_TYPE_SRP_KEY_PAIR_BASE          ((psa_key_type_t) 0x7700)
+#define PSA_KEY_TYPE_SRP_PUBLIC_KEY_BASE        ((psa_key_type_t) 0x4700)
+#define PSA_KEY_TYPE_SRP_GROUP_MASK             ((psa_key_type_t) 0x00ff)
+
+ /** SRP key pair. Both the client and server key.
+ *
+ * The size of a SRP key is the size associated with the Diffie-Hellman
+ * group. See the documentation of each Diffie-Hellman group for details.
+ * To construct a SRP key pair, the password hash must be imported.
```
athoelke, line 15: Later, the documentation talks about constructing an SRP-6 key as output from a password-hash type of KDF, rather than by using `psa_import_key()` on some data.

athoelke (changes requested):
Thank you for the contribution!

Some further clarification of the definition of the protocol, as the public documentation has various definitions for some of the computational steps. This is necessary to ensure compatible implementations of algorithms in the Crypto API.

If we need additional flexibility for different uses of SRP-6, then we will need additional algorithm identifiers, or other parameterisation of the PAKE cipher-suite.

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: Agreed, it should be:

"To construct an SRP key pair, it must either be output from a key derivation operation, or imported."

(similar to SPAKE2+)

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: This is SRP-6a as defined by RFC 5054 including errata `(k = Hash(N || PAD(g))`.

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: Agreed.

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: Yes, flexibility is important for that reason.

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: We would suggest that both roles call `psa_pake_set_user` (`psa_pake_set_peer` is not required).

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: The variant where M1 = H(H(N) XOR H(g) | H(U) | s | A | B | K) and M2 = H(A | M1 | K) seems to be the clear standard.
It is defined or used by:

- RFC2945
- RFC5054 (by referring to RFC2945)
- srp.stanford.edu/design (for SRP-3, SRP-6, and SRP-6a)
- The Stanford reference implementation
- The reference implementation on Wikipedia
- The SRP test vectors on github.com/secure-remote-password/test-vectors
- HomeKit (by referring to the Stanford reference implementation)
- The Wolfcrypt library

(see also the comment for the client proof phase call)

headers/crypto/1.2/psa/crypto-srp.h
oberon-sk: The variant where M1 = H(H(N) XOR H(g) | H(U) | s | A | B | K) and M2 = H(A | M1 | K) seems to be the clear standard.
It is defined or used by:

- RFC2945
- RFC5054 (by referring to RFC2945)
- srp.stanford.edu/design (for SRP-3, SRP-6, and SRP-6a)
- The Stanford reference implementation
- The reference implementation on Wikipedia
- The SRP test vectors on github.com/secure-remote-password/test-vectors
- HomeKit (by referring to the Stanford reference implementation)
- The Wolfcrypt library

(see also the comment for the server proof phase call)

headers/crypto/1.2/psa/crypto-srp.h
athoelke: Some descriptions of SRP-6 show the client private key being the output of a simple hash - so we need to ensure the correct construction of an SRP-6 key from an imported data buffer (the hash output), as well as construction via a KDF (such as the original SRP specification double-hash algorithm).

For a general KDF, how does `psa_key_derivation_output_key()` know how much data to extract when deriving the SRP-6 key? The KDF presented in the SRP definition only outputs the hash output size. What about KDFs that do not have limited output? - we can either define the behavior explicitly, or recommend the application extract the required number of KDF bytes, and then import the SRP key?

oberon-sk: This is a general problem for the derivation of structured keys. The PSA core has to ‘know’ somehow how much data the driver needs to construct the key. The “[psa-driver-interface.md](https://github.com/Mbed-TLS/mbedtls/blob/v3.5.2/docs/proposed/psa-driver-interface.md)” document mentions this under “[Open questions around cooked key derivation](https://github.com/Mbed-TLS/mbedtls/blob/v3.5.2/docs/proposed/psa-driver-interface.md#open-questions-around-cooked-key-derivation)”. The idea of an additional entry point to determine the expected length mentioned there is possibly the best solution.
For trial-and-error type key derivation like FFDH the driver simply returns `PSA_ERROR_INSUFFICIENT_DATA` to signal the core to call it again with the next chunk of data.

Issue #95, opened 2023-08-31
Define API for Module-Lattice-based Key-Encapsulation Mechanism (ML-KEM aka Crystals-KYBER)


athoelke:
**Updated 02-09-2024:** The final version of the specification is now published https://csrc.nist.gov/pubs/fips/203/final.

NIST has now published a draft specification for a Key Encapsulation algorithm based on Crystals-KYBER. The algorithm is designated ML-KEM, and will be published as FIPS 203. The draft (open for review until 22 Nov 2023) can be downloaded from https://csrc.nist.gov/pubs/fips/203/ipd.

NIST is planning to standardize three parameterized variants of ML-KEM: ML-KEM-512, ML-KEM-768, ML-KEM-1024, which provide increasing levels of security for increasing computation and size of keys and encapsulated key values.

The Crypto API should define an API for using this algorithm.

athoelke:
### API design thoughts

On first review of the specification, it might be tempting to try and implement the public-key encryption algorithm, K-PKE, that underlies ML-KEM, as a PSA Crypto API algorithm. K-PKE looks similar to other Crypto API asymmetric encryption algorithms, except that it requires an additional input of 512 bits of random data for the encryption step (this is required to support the ML-KEM decapsulation verification). Additionally, FIPS 203 states that K-PKE must not be used for asymmetric encryption:

> **K-PKE is only a component**. The public-key encryption scheme K-PKE described in Section 5 **shall not** be used as a stand-alone cryptographic scheme. Instead, the algorithms that comprise K-PKE may only be used as subroutines in the algorithms of ML-KEM. In particular, the algorithms K-PKE.KeyGen (Algorithm 12), K-PKE.Encrypt (Algorithm 13), and K-PKE.Decrypt (Algorithm 14) are **not approved** for use as a public-key encryption scheme.

It seems technically possible to just define K-PKE (and the additional Hash and XOF functions) at the Crypto API layer, and leave the ML-KEM wrapper over K-PKE to the application layer; and so avoid introducing a new API category for KEMs. This would still require new API for asymmetric-encryption-with-context, and standardizing how to provide XOF algorithms via the API. However, this puts a significant responsibility for secure implementation of ML-KEM into the application.

Alternatively, we can introduce a new category of algorithm, "key encapsulation", along with associated key types, that is designed to directly support the ML-KEM operations. In anticipation of other KEMs being specified in future, this is probably the better approach.
Note the proposed SP 800-227 document, that:

[...]

PR #182, opened 2024-03-07
Encapsulation - ECIES

MarcusJGStreets:
Added key encapsulation functions as a precursor for PQC.
So far the only method is ECIES
Need to confirm the value for the ECIES macro

### DIFF
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -232,7 +232,6 @@ The usage flags are encoded in a bitmask, which has the type `psa_key_usage_t`. 
 
     This flag must be present on keys used with the following APIs:
 
-    *   `psa_key_agreement()`
     *   `psa_key_derivation_key_agreement()`
     *   `psa_raw_key_agreement()`
 
@@ -250,20 +249,29 @@ The usage flags are encoded in a bitmask, which has the type `psa_key_usage_t`. 
 
     If this flag is present on all keys used in calls to `psa_key_derivation_input_key()` for a key derivation operation, then it permits calling `psa_key_derivation_verify_bytes()` or `psa_key_derivation_verify_key()` at the end of the operation.
 
-
-.. macro:: PSA_KEY_USAGE_ENCAPSULATE_KEY
-    :definition: ((psa_key_usage_t)0x00008000)
+.. macro:: PSA_KEY_USAGE_ENCAPSULATE
+    :definition: ((psa_key_usage_t)0x00010000)
 
     .. summary::
         Permission to encapsulate new keys.
 
-    This flag is required to use a private key to encapsulate new keys to send to a counter party, or to decapsulate a key buffer obtained from a counter party.
+    This flag is required to encapsulate new keys to send to a counter party.
+
+    This flag must be present on public keys used with the following APIs:
+
+    *   `psa_encapsulate()`
+
+.. macro:: PSA_KEY_USAGE_DECAPSULATE
+    :definition: ((psa_key_usage_t)0x00020000)
+
+    .. summary::
+        Permission to decapsulate an encapsulated key.
+
+    This flag is required to decapsulate a key buffer obtained from a counter party.
 
     This flag must be present on private keys used with the following APIs:
 
-    *   `psa_encapsulate_key()`
-    *   `psa_decapsulate_key()`
-
+    *   `psa_decapsulate()`
 
 
 .. function:: psa_set_key_usage_flags
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -33,10 +33,10 @@ The following algorithm policies are supported:
 *   `PSA_ALG_NONE` does not permit any cryptographic operation with the key. The key can still be used for non-cryptographic actions such as exporting, if permitted by the usage flags.
 *   A specific algorithm value permits exactly that particular algorithm.
 *   A signature algorithm constructed with `PSA_ALG_ANY_HASH` permits the specified signature scheme with any hash algorithm. In addition, :code:`PSA_ALG_RSA_PKCS1V15_SIGN(PSA_ALG_ANY_HASH)` also permits the `PSA_ALG_RSA_PKCS1V15_SIGN_RAW` signature algorithm.
-*   A raw key agreement algorithm also permits the specified key agreement scheme to be combined with any key derivation algorithm.
+*   A standalone key agreement algorithm also permits the specified key agreement scheme to be combined with any key derivation algorithm.
 *   An algorithm built from `PSA_ALG_AT_LEAST_THIS_LENGTH_MAC()` permits any MAC algorithm from the same base class (for example, CMAC) which computes or verifies a MAC length greater than or equal to the length encoded in the wildcard algorithm.
 *   An algorithm built from `PSA_ALG_AEAD_WITH_AT_LEAST_THIS_LENGTH_TAG()` permits any AEAD algorithm from the same base class (for example, CCM) which computes or verifies a tag length greater than or equal to the length encoded in the wildcard algorithm.
-
+*   The `PSA_ALG_CCM_STAR_ANY_TAG` wildcard algorithm permits the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm, the `PSA_ALG_CCM` AEAD algorithm, and the :code:`PSA_ALG_AEAD_WITH_SHORTENED_TAG(PSA_ALG_CCM, tag_length)` truncated-tag AEAD algorithm for ``tag_length`` equal to 4, 8 or 16.
 When a key is used in a cryptographic operation, the application must supply the algorithm to use for the operation. This algorithm is checked against the key's permitted-algorithm policy.
 
 .. function:: psa_set_key_algorithm
@@ -231,11 +231,12 @@ The usage flags are encoded in a bitmask, which has the type `psa_key_usage_t`. 
 
     This flag must be present on keys used with the following APIs:
 
+    *   `psa_key_agreement()`
     *   `psa_key_derivation_key_agreement()`
     *   `psa_raw_key_agreement()`
 
-    If this flag is present on all keys used in calls to `psa_key_derivation_input_key()` for a key derivation operation, then it permits calling `psa_key_derivation_output_bytes()` or `psa_key_derivation_output_key()` at the end of the operation.
-
+    If this flag is present on all keys used in calls to `psa_key_derivation_input_key()` for a key derivation operation, then it permits calling `psa_key_derivation_output_bytes()`, `psa_key_derivation_output_key()`, `psa_key_derivation_verify_bytes()`, or `psa_key_derivation_verify_key()` at the end of the operation.
+    
 .. macro:: PSA_KEY_USAGE_VERIFY_DERIVATION
     :definition: ((psa_key_usage_t)0x00008000)
 
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -250,6 +250,22 @@ The usage flags are encoded in a bitmask, which has the type `psa_key_usage_t`. 
 
     If this flag is present on all keys used in calls to `psa_key_derivation_input_key()` for a key derivation operation, then it permits calling `psa_key_derivation_verify_bytes()` or `psa_key_derivation_verify_key()` at the end of the operation.
 
+
+.. macro:: PSA_KEY_USAGE_ENCAPSULATE_KEY
+    :definition: ((psa_key_usage_t)0x00008000)
+
+    .. summary::
+        Permission to encapsulate new keys.
+
+    This flag is required to use a private key to encapsulate new keys to send to a counter party, or to decapsulate a key buffer obtained from a counter party.
+
+    This flag must be present on private keys used with the following APIs:
+
+    *   `psa_encapsulate_key()`
+    *   `psa_decapsulate_key()`
+
+
+
 .. function:: psa_set_key_usage_flags
 
     .. summary::
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -33,10 +33,9 @@ The following algorithm policies are supported:
 *   `PSA_ALG_NONE` does not permit any cryptographic operation with the key. The key can still be used for non-cryptographic actions such as exporting, if permitted by the usage flags.
 *   A specific algorithm value permits exactly that particular algorithm.
 *   A signature algorithm constructed with `PSA_ALG_ANY_HASH` permits the specified signature scheme with any hash algorithm. In addition, :code:`PSA_ALG_RSA_PKCS1V15_SIGN(PSA_ALG_ANY_HASH)` also permits the `PSA_ALG_RSA_PKCS1V15_SIGN_RAW` signature algorithm.
-*   A standalone key agreement algorithm also permits the specified key agreement scheme to be combined with any key derivation algorithm.
+*   A raw key agreement algorithm also permits the specified key agreement scheme to be combined with any key derivation algorithm.
 *   An algorithm built from `PSA_ALG_AT_LEAST_THIS_LENGTH_MAC()` permits any MAC algorithm from the same base class (for example, CMAC) which computes or verifies a MAC length greater than or equal to the length encoded in the wildcard algorithm.
 *   An algorithm built from `PSA_ALG_AEAD_WITH_AT_LEAST_THIS_LENGTH_TAG()` permits any AEAD algorithm from the same base class (for example, CCM) which computes or verifies a tag length greater than or equal to the length encoded in the wildcard algorithm.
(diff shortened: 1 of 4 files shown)


athoelke:
Considering this API design in light of what we will need for ML-KEM (FIPS 203 - see #95): these might require different APIs?

The protocol for ML-KEM looks a bit like this:

image

Here, the key-pair is used something vaguely like a PKE key-pair: Bob uses Alice's encapsulation (public) key to encrypt some information that can be used to derive a shared secret. Alice can decrypt that information to derive the same secret. In this exchange, Bob could use a copy of Alice's k<sub>encaps</sub> provided in a trusted certificate, thus ensuring that only the real Alice (who has protected k<sub>decaps</sub>, of course) will be able to correct derive the shared secret. But there is no authentication of Bob.

There is no concept of both parties having their own pair of keys, and using both their own private and the other participants public key for the encapsulation and decapsulation procedures - which is what the API for ECIES does above. 

So, it does not seem possible that we can use the same API for ECIES and ML-KEM?

athoelke:
I've had a more detailed read-through of the ECIES definition (from [SEC1](https://www.secg.org/sec1-v2.pdf) §5.1), and it seems that the shared secret construction is more similar to ML-KEM than I first thought.

Note that the proposal does not implement the whole of ECIES, purely the key establishment mechanism. The full definition of ECIES requires the participants to agree on a suite of parameters and choices with respect to encryption and authentication of the message to transmit (e.g. *SharedInfo<sub>1/2</sub>*, choice of encryption and MAC algorithms, etc.). In keeping with the low-level design of the Crypto API, this seems appropriate: the application can use the appropriate key derivation, encryption and MAC operations as required.

Summary of ECIES:
1. In ECIES, one participant (A) creates a long term key pair, and publishes the public key (or provides it to the other participant, B).
2. When transmitting an encrypted message to A, B uses A's public key and an ephemeral key-pair to create the shared secret - from which the keys for encrypting and authenticating the message are derived.
3. B then sends the ephemeral public key, the encrypted message, and the authentication tag to A.
4. A can then also construct the shared secret to verify and decrypt the message.

Here is a generalised version of the sequence diagram above:

image

Then for ECIES:
* k<sub>encaps</sub> and k<sub>decaps</sub> are the public and private ECC keys for Alice.
* When Bob invokes Encaps() using k<sub>encaps</sub>, this creates the ephemeral key pair, computes the ECDH shared secret, and returns the ephemeral public key as the *encaps-text*, and the ECDH result as K<sub>shared</sub>.
* When Alice invokes Decaps() with the provided *encaps-text* from Bob, and k<sub>decaps</sub>, this copmutes the ECDH shared secret and returns this as K<sub>shared</sub>.
* The use of the shared secret to construct keys for encryption and authentication of the message, or verifying and decrypting the ciphertext is up to the applications.

Thus, if we could adjust the proposed API to align it with what ML-KEM would require, the API could be common for the two algorithms:
* `psa_encapsulate_key()` does not take a private key, but only a public key and generates the ephemeral key pair, providing the public part as the output buffer
* `psa_decapsulate_key()` takes a private key, and the buffer (sent from the other participant), and decodes the buffer as the other party's public key.

If this makes sense, then `psa_encapsulate_key()` will need to take the public key as a key handle (not as a key data buffer) - as this indicates the EC group over which the ECDH computation must happen.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,317 @@
+Key Encapsulation
+=================
+
+Elliptic Curve Integrated Encryption Scheme
+-------------------------------------------
+
+The Elliptic Curve Integrated Encryption Scheme was fist proposed by Shoup, then imporved by Ballare and Rogaway.
```
sevar, line 7: typo: imporved -> improved

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,317 @@
...
+    .. param:: const psa_key_id_t * counterparty_key
+        Public key of the peer. The peer key must be in the same format that `psa_import_key()` accepts for the public key type corresponding to the type of ``private_key``. That is, this function performs the equivalent of :code:`psa_import_key(..., peer_key, peer_key_length)`, with key attributes indicating the public key type corresponding to the type of ``private_key``. For example, for ECC keys, this means that peer_key is interpreted as a point on the curve that the private key is on. The standard formats for public keys are documented in the documentation of `psa_export_public_key()`.
+
+    .. param:: const psa_key_id_t * private_key
+        Identifier of the key belong to the person performing the encapsulation. 
+        It must be an asymmetric key pair. 
+        The key must permit the usage `PSA_KEY_USAGE_ENCAPSULATE_KEY`
+
+    .. param:: psa_algorithm_t alg
+        The ful encapsulation algorithm to use: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_ENCAPSULATION(alg)` is true and :code:`PSA_ALG_IS_RAW_ENCAPSULATION(alg)` .
```
sevar, line 83: ful -> full

sevar (commented):
minor typos

sevar:
From API perspective, KEM is very similiar to PKE (like ECIES or RSA.PKE), wouldn't be sufficient to extend the API around ``psa_asymmetric_encrypt`` and ``psa_asymmetric_decrypt`` to support all, ECIES, ML-KEM and FrodoKEM? The only problem seems to be with keys, ML-KEM in opposition to other schemes, consider ``decapsulation key`` as both, secret and public key due to the FO-Transformation, but due to usage of key handler in the API, only key_attributes needs to be set properly, no?

athoelke:
> From API perspective, KEM is very similiar to PKE (like ECIES or RSA.PKE), wouldn't be sufficient to extend the API around `psa_asymmetric_encrypt` and `psa_asymmetric_decrypt` to support all, ECIES, ML-KEM and FrodoKEM?

There is a difference in the application expectation:

* With `psa_asymmetric_encrypt()` the application supplies data to encrypt for transfer to the recipient.
* KEM internally generates the shared secret (or its seed), and encrypts (encapsulates) that for transfer, there is no application data involved at this step.

From an application perspective, ECIES looks similar to asymmetric encryption, but uses public-key cryptography only to establish a shared secret for use in encryption of the message. This is difficult to encode within the PSA API for asymmetric encryption, as the algorithm specification has to include the the KDF used to derived the encryption/MAC keys from the shared secret, the cipher algorithm, and the MAC algorithm. Combined key agreement with key derivation we already have, but we cannot also flexibly specify the cipher and MAC components required for ECIES in a single algorithm id.

However, a lower level API for ECIES seems to fit well with the PSA Crypto API design, as suggested above. This provides the flexibility for application selection of cipher suites for the encryption/authentication, and aligns with the behavior of KEM algorithms.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+           *   ``PSA_ECC_FAMILY_SECP_XX``
+           *   `PSA_ECC_FAMILY_FRP`
+           *   `PSA_ECC_FAMILY_BRAINPOOL_P_R1`
+           *   `PSA_ECC_FAMILY_MONTGOMERY`
+
+.. macro:: PSA_ALG_ENCAPSULATION
+    :definition: /* specification-defined value */
+
+    .. summary::
+        Macro to build a combined algorithm that chains a key encapsulation with a key derivation.
```
athoelke, line 44: I don't think it will be easy to drive a combined algorithm here. The KDF requires additional inputs, and the specific input steps are KDF specific.

For the PSA Crypto API, I think it might be best to just provide an encapsulation API that results in the shared secret - for example, the value denoted `Z` in the SEC1 specification of ECIES. Derivation of the secret into encryption and authentication keys is then left to the application - or a higher-level ECIES function that wraps the PSA Crypto API calls.

This places these APIs at the same level as the key agreement APIs - typically an application must use key derivation to convert a possibly-biased shared secret into a usable encryption key.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+.. _encapsulation-algorithms:
+
+Encapsulation Algorithms
+------------------------
+.. function:: psa_encapsulate_key
+
+    .. summary::
+        Generate a new key pair and a use that to encapsulate a new symmetric key, emitting it both as a key object and an encapsulation to send to a counter party along with the public key from the ephemeral key pair.
+
+    .. param:: const psa_key_id_t * counterparty_key
```
athoelke, line 73: `psa_key_id_t` is a simple type, for input parameters, we pass these by value.

Given the comments about the other parameters below - I think this could just be named `key`.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+        The identifier for the public key of the peer. You must have previously imported this key using `psa_import_key()`, and specified the key attributes for the public key type corresponding to the type required for the encapsulation, and the usage usage `PSA_KEY_USAGE_ENCAPSULATE_KEY`.
+
+    .. param:: psa_key_id_t * key_pair
```
athoelke, line 76: There is no need to retain or return the ephemeral key pair used during the encapsulation. For ECIES, the private part is already used to create the shared secret key that is returned as `output_key`, the public part is returned in the `encapsulation` output.

For ML-KEM, the mechanism is entirely different, and there is no ephemeral key pair.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+        On success an identifier for the new key pair, which was used to perform the encapsulation.  
+        The key will be the key type specified in the atributes of the counterparty key. The private key will be given the usage `PSA_KEY_USAGE_ENCAPSULATE_KEY`. The public key only will be given the usage `PSA_KEY_USAGE_EXPORT`.  
+        `PSA_KEY_ID_NULL` on failure
+
+    .. param:: uint8_t * ephemeral_public_key
```
athoelke, line 81: Not required, as already provided in `encapsulation`.

The ephemeral public key in ECIES is the encapsulation data.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+        
+    .. param:: size_t ephemeral_public_key_size
+        Size of the ``ephemeral_public_key`` buffer in bytes.
+        This must be at least :code:`PSA_EXPORT_PUBLIC_KEY_OUTPUT_SIZE(alg)`.
+        A buffer of at least :code:`PSA_EXPORT_PUBLIC_KEY_MAX_OUTPUT_SIZE`. is guaranteed not to fail due to buffer size for any supported encapsulation algorithm.
+
+    .. param:: psa_algorithm_t alg
+        The ful encapsulation algorithm to use: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_ENCAPSULATION(alg)` is true and :code:`PSA_ALG_IS_RAW_ENCAPSULATION(alg)` is false .
+
+    .. param:: const psa_key_attributes_t * attributes
```
athoelke, line 92: I suspect that this should probably have the same constraints/documentation as `psa_key_agreement()`. Depending on the algorithm, the output key/secret may be biased, and the size of the key is determine wholly by the algorithm and key parameters.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+
+        .. note::
+            This is an input parameter: it is not updated with the final key attributes. The final attributes of the new key can be queried by calling `psa_get_key_attributes()` with the key's identifier.
+        
+
+    .. param:: psa_key_id_t * output_key
+        On success, an identifier for the newly created key. `PSA_KEY_ID_NULL` on failure.
+        
+    .. param:: uint8_t * encapsulation
+        Buffer where the encapsulated key is to be written, ready to be sent to the counterparty.
```
athoelke, line 110: In general, the encapsulation data might contain a key (as for ECIES), but might contain entirely different data (as for ML-KEM).

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+
+.. function:: psa_decapsulate_key
+
+    .. summary::
+        Uses a private key to decapsulate an encapsulation received from a counter party. 
+
+    .. param:: const psa_key_id_t * counterparty_key
+        Public key of the peer. The peer key must be in the same format that `psa_import_key()` accepts for the public key type corresponding to the type of ``private_key``. That is, this function performs the equivalent of :code:`psa_import_key(..., peer_key, peer_key_length)`, with key attributes indicating the public key type corresponding to the type of ``private_key``. For example, for ECC keys, this means that peer_key is interpreted as a point on the curve that the private key is on. The standard formats for public keys are documented in the documentation of `psa_export_public_key()`.
+
+    .. param:: const psa_key_id_t * private_key
```
athoelke, line 160: For this operation, this can probably just be `key`.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+        Buffer containing the encapsulated key that was received from the counterparty.
+        
+    .. param:: size_t encapsulation_size
+        Size of the ``encapsulation`` buffer in bytes.
+
+
+    .. param:: psa_algorithm_t alg
+        The encapsulation algorithm to use: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_ENCAPSULATION(alg)` is true.
+
+    .. param:: const psa_key_attributes_t * attributes
```
athoelke, line 175: The same comments apply as the attributes parameter to `psa_encapsulate_key()`.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+
+    .. subsection:: Compatible key types
+
+        The resulting combined encapsulation algorithm is compatible with the same key types as the raw encapsulation algorithm used to construct it.
+
+.. _encapsulation-algorithms:
+
+Encapsulation Algorithms
+------------------------
+.. function:: psa_encapsulate_key
```
athoelke, line 68: Should the API name perhaps be `psa_encapsulate()`? - and likewise for `psa_decapsulate()`.

E.g. for ML-KEM, the encapsulation is not a key.

doc/crypto/api/ops/encap.rst
```diff
@@ -0,0 +1,326 @@
...
+-------------------------------------------
+
+The Elliptic Curve Integrated Encryption Scheme was fist proposed by Shoup, then improved by Ballare and Rogaway.
+
+The original specification permitted a number of variants. This specification only defines the version specified in [SEC1], that is with the use of labels and with the label size defined in bytes. 
+
+It is possible that some applications may need to use older versions to interoperate with legacy systems. 
+While the application can always implement this using the other algorithm functions provided, however, an implementation may choose to add these as a convenience in the implementation numbering space. 
+
+.. macro:: PSA_ALG_ECIES
```
athoelke, line 14: Do we want/need to provide separate algorithms for ECIES with ECDH, and for ECIES with EC-Cofactor-DH? (as described in SEC1). I assuming that this cannot be deduced from the key attributes.

athoelke (changes requested):
Although this is much closer to what I think would work well, there is some ECIES-specific aspects that we probably don't need and would prevent reuse for ML-KEM:

* The encapsulation function does not have to provide an outputs related to the ephemeral keys. The encapsulation data output does include the public key, but that is just an property of this specific algorithm.
* Combining an arbitrary KDF introduces the complexity of specifying the other inputs. Better to copy the pattern of `psa_key_agreement()` and output a possibly-biased shared secret that the caller can input to a KDF of their choosing.
* Smaller issues about naming, and having two flags for this asymmetrical API.

doc/crypto/api/keys/policy.rst
```diff
@@ -250,6 +248,22 @@ The usage flags are encoded in a bitmask, which has the type `psa_key_usage_t`.
 
     If this flag is present on all keys used in calls to `psa_key_derivation_input_key()` for a key derivation operation, then it permits calling `psa_key_derivation_verify_bytes()` or `psa_key_derivation_verify_key()` at the end of the operation.
 
+
+.. macro:: PSA_KEY_USAGE_ENCAPSULATE_KEY
```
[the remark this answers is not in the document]
MarcusJGStreets: Probably right - changed here and in encap.rst

doc/crypto/api/ops/encap.rst
MarcusJGStreets: PSA_ALG_ECIES references Sec 1 so is cofactor. 

We possibly should call it PSA_ALG_ECIES_SEC1

We could add PSA_ALG_ECIES_18033 to cover the ISO version without cofactors. 
However, in earlier discussion in the Tuesday calls, Giles suggested we just define one for now.

It would be trivial to add PSA_ALG_ECIES_18033

athoelke: SEC1 v2.0 describes ECIES using either ECDH or EC-Cofactor-DH. See §5.1.1 point 4 (page 51 in https://www.secg.org/sec1-v2.pdf), which has this choice as a selected option/parameter by the participants.

However, `PSA_ALG_ECDH` in the Crypto API is described as implementing the cofactor computation. See the last few paragraphs in the [description of `PSA_ALG_ECDH`](https://arm-software.github.io/psa-api/crypto/1.2/api/ops/ka.html#c.PSA_ALG_ECDH). So it is reasonable for `PSA_ALG_ECIES` to be defined as using the same key agreement computation.

Issue #167, opened 2024-02-12
Custom methods for key generation and key derivation


gilles-peskine-arm:
The PSA API should provide ways to customize how key material is created from random or pseudorandom inputs, i.e. key generation and key derivation. The scope here is for customizing the construction of keys (especially cooked keys, i.e. keys that aren't just chosen uniformly among bit-strings of a given length). Here are some possible use cases:

* Generate an RSA key with a chosen public exponent. The current API imposes e=65537. There is demand for e=3, and occasionally for other values. This is also relevant for derivation, although the PSA API does not specify how to do derivation for RSA.
* Choose among several plausible methods for deriving an RSA key. This would likely remain implementation-specific, since the PSA API does not specify how to do derivation for RSA.
* When deriving an ECC key (which involves randomly selecting an integer in a range that isn't a power of 2), use the fixed-input-size-divided-by-N method rather than the method imposed by the API, which is to read input repeatedly until it's in range.
* Choose a custom random generator when generating a key randomly.

Urgency: Mbed TLS is interested in having a way to generate an RSA key with a chosen public exponent soon. Ideally we would like to include a beta version of this in our next release (code freeze: mid-March), but we don't know whether we'll have time to do the coding even.

My general idea is to define new functions `psa_generate_key_ext()` and `psa_key_derivation_output_key_ext()` (tentative names), which are similar to `psa_generate_key()` and `psa_key_derivation_output_key()` but take an extra parameter indicating the custom generation/derivation method. The type of the method parameter has a default value which means the same thing that the non-ext functions do.

gilles-peskine-arm:
See https://github.com/Mbed-TLS/mbedtls/pull/8815 for an API proposal (with a prototype implementation).

gilles-peskine-arm:
On Tuesday 13 Feb 2024, we decided to go ahead with the design in https://github.com/Mbed-TLS/mbedtls/pull/8815 at https://github.com/Mbed-TLS/mbedtls/pull/8815/commits/d8ba2b8716a1854c51b94ac69c24317b5ab65409 with the change from passing `method_length` to passing `method_data_length` as proposed in https://github.com/Mbed-TLS/mbedtls/pull/8815#discussion_r1487520828.

gilles-peskine-arm:
[...]

Issue #149, opened 2024-01-16
Importing and exporting keys that are in non-default formats


athoelke:
_This issue is a broader set of use cases than the one defined in #44, which only considers the import of a key from a data format that specifies the key type and/or policy._

There are a number of uses cases where built-in support for additional key data formats are valuable for applications using the Crypto API:

* Importing data in a format that provides some or all of the key meta-data alongside the key material. (see also #44)
* Exporting keys to a standard format, that is not the default key format for the API.
* Import and export of wrapped keys, that use a device-specific, secret wrapping key and an implementation-specified algorithm. This use case can be useful for provisioning keys to a device, or securely storing keys outside of the key-store.

For more general key-wrapping support, where the application can select the algorithm and key-wrapping key, see #50.

athoelke:
Initial work drafting an API to support the first use case - "import keys in other formats" - has been carried out in the Mbed TLS project, here: https://github.com/Mbed-TLS/mbedtls/pull/7910.

I plan to extend that design to support key export, and implementation-defined formats, in order to support the other use cases.

athoelke:
### Key formats to support

Key format related requirements:

* Can specify the current default key format in the Crypto API, for all key types.
* Can specify values in the Mbed TLS draft, for asymmetric RSA and ECC keys:

  * DER or PEM encoded RSAPublicKey, as defined in RFC 8017
  * DER or PEM encoded SubjectPublicKeyInfo data structure, as defined in RFC 5280
  * DER or PEM encoded RSAPrivateKey, as defined in RFC 8017
  * DER or PEM encoded ECPrivateKey, as defined in RFC 5915
  * DER or PEM encoded OneAsymmetricKey (previously PrivateKeyInfo), as defined in RFC 5958

  The Mbed TLS draft proposes that the SubjectPublicKeyInfo format can also be used to import key data that is in one of the enclosed structure formats, in particular, RSAPublicKey. Likewise, the OneSymmetricKey format also permits key data in the RSAPrivateKey or ECPrivateKey format.

  Is this relaxation important for the key import API?

* An implementation can define implementation-specific formats.
* Are there other important key formats that should be supported?

The API design in https://github.com/Mbed-TLS/mbedtls/pull/7910 only addresses key import, and the definition of key formats in the `psa_key_data_format_t` enumeration is imprecise: each of the five proposed key format values support both PEM and DER encoding. For export, the encoding has to be specified by the application.

[...]

PR #184, opened 2024-03-13
Reorganise key format and key derivation information

### LINKED ISSUE #183: Mistaken assumption in HMAC key definition and export format
While reworking the key formats and derivation methods (see the open issue in #177), I have noticed a long-term anomaly in the specification. HMAC keys describe being truncated on import, based on the block size of the associated hash algorithm. See [`PSA_KEY_TYPE_HMAC`](https://arm-software.github.io/psa-api/crypto/1.2/api/keys/types.html#c.PSA_KEY_TYPE_HMAC), and also the entry for HMAC keys in [*Key formats*](https://arm-software.github.io/psa-api/crypto/1.2/api/keys/management.html#key-formats).

The latter is written as if the HMAC key is parameterised:

> ### `PSA_KEY_TYPE_HMAC` (hash_alg)
>
> For HMAC keys that are shorter than, or equal in size to, the block size of `hash_alg`, the format is the raw bytes of the key.
>
>HMAC keys that are longer than the block size of `hash_alg`, are permitted in a call to `psa_import_key()`. For such a key, the output of `psa_export_key()` is an implementation defined choice between the following:
>
> * The raw bytes of the key.
> * The raw bytes of the hash of the key, using hash_alg.

This is clearly wrong. Only the HMAC **_algorithm_** is parameterised.

athoelke:
_[Updated: rebased after publication of Crypto API 1.2.1. Updated the rendered PDF]_

This is a follow-up to #177, which moves the information about key formats, and key derivation, into each key type definition.

This is more user-centric than providing all the key formats in a single place, and all the key derivation methods in another single place. It will also scale better when we add additional formats, or key construction methods (see #149 and #167).

The new commits (since #177) start at https://github.com/ARM-software/psa-api/pull/184/commits/0733f8164fa650b3e1f128963e44192e48e4b9fa

Preview of the PDF: [IHI0086-PSA_Certified_Crypto_API-1.3.0-key-formats-draft.2.pdf](https://github.com/ARM-software/psa-api/files/14791577/IHI0086-PSA_Certified_Crypto_API-1.3.0-key-formats-draft.2.pdf)


athoelke:
I'm not sure if the 'Key format' and 'Key derivation' sections might benefit from some boilerplate text to indicate the sub-section's relevance to `psa_import_key()`/`psa_export_key()` and `psa_key_derivation_ouput_key()` respectively?

athoelke:
Rebased after #177 merged (with minor fixes)

PR #185, opened 2024-03-14
Remove dependency on the associated hash from the definition of HMAC keys

### LINKED ISSUE #183: Mistaken assumption in HMAC key definition and export format

athoelke:
Fixes #183

### DIFF
--- a/doc/crypto/api/keys/management.rst
+++ b/doc/crypto/api/keys/management.rst
@@ -549,16 +549,6 @@ This section defines the format of the key data that an implementation is requir
 
             The key data is the concatenation of the two or three DES keys.
 
-    *   -   HMAC
-        -   :code:`PSA_KEY_TYPE_HMAC(hash_alg)`
-
-            For HMAC keys that are shorter than, or equal in size to, the block size of ``hash_alg``, the format is the raw bytes of the key.
-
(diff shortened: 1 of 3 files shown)


MarcusJGStreets:
Looks OK to me.

PR #186, opened 2024-03-14
Improve the description of using IVs in PSA_ALG_CTR

### LINKED ISSUE #181: crypto: problematic specification of IV for PSA_ALG_CTR
The specification of [`PSA_ALG_CTR`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/ciphers.html#c.PSA_ALG_CTR) has inconsistent provisions regarding the counter block.

CTR mode does not have a single specification. It is really a family of modes that arrange for the counter value uniqueness in different ways.

> During the counter block update operation, the counter block is treated as a single big-endian encoded integer and the update operation increments this integer by 1.

This is one way to do it, and it's what Mbed TLS does.

> A counter block value must only be used once across all messages encrypted using the same key value. This is typically achieved by splitting the counter block into a nonce, which is unique among all message encrypted with the key, and a counter which is incremented for each block of a message.

This is not how it works: it's only compatible with the paragraph cited above if the none part is 0 bytes, but then the nonce part is not a nonce since reuse is allowed (as long as the resulting counter value is unique across blocks).

> For example, when using AES-CTR encryption, which uses a 16-byte block, the application can provide a 12-byte nonce when setting the IV. This leaves 4 bytes for the counter, allowing up to 2^32 blocks (64GB) of message data to be encrypted in each message.

This example does not comply to the specification above. Note that the length passed to `set_iv` is irrelevant to the internal operation of the cipher.

Also:

> A call to [`psa_cipher_set_iv()`](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/ciphers.html#c.psa_cipher_set_iv) on a multi-part cipher operation requires an IV that is between 1 and n bytes in length, where n is the cipher block length. The counter block is initialized using the IV, and padded with zero bytes up to the block length.

We might as well allow a 0-byte IV to start the counter at 0.

athoelke:
Fixes #181

[As we are patching 1.2 to fix some minor issues, it is worth including this clarification before 1.3.]

### DIFF
--- a/doc/crypto/api/ops/ciphers.rst
+++ b/doc/crypto/api/ops/ciphers.rst
@@ -115,29 +115,30 @@ Cipher algorithms
 
     The CTR block cipher mode is defined in :cite-title:`SP800-38A`.
 
-    CTR mode requires a *counter block* which is the same size as the cipher block length. The counter block is updated for each block (or a partial final block) that is encrypted or decrypted.
+    CTR mode operates using a *counter block* which is the same size as the cipher block length. The counter block is updated for each block, or a partial final block, that is encrypted or decrypted.
 
-    A counter block value must only be used once across all messages encrypted using the same key value. This is typically achieved by splitting the counter block into a nonce, which is unique among all message encrypted with the key, and a counter which is incremented for each block of a message.
+    For the `PSA_ALG_CTR` algorithm, the counter block is initialized from the IV. The counter block is then treated as a single, big-endian encoded integer, and the counter block is updated by incrementing this integer by ``1``.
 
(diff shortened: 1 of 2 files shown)


MarcusJGStreets (dismissed):
You have lost the sentence 

A counter block value must only be used once across all messages encrypted using the same key value.

This is the key cryptographic property, so should be retained.

athoelke:
> You have lost the sentence
> 
> A counter block value must only be used once across all messages encrypted using the same key value.
> 
> This is the key cryptographic property, so should be retained.

I've added a patch to reinstate it (or an equivalent). Please check the result conveys the intended meaning effectively.

PR #195, opened 2024-03-28
Add a macro to evaluate the maximum asymmetric key export size

### LINKED ISSUE #180: crypto: MAX_SIZE constant for public key or key pair
The PSA crypto API has constants `PSA_EXPORT_KEY_PAIR_MAX_SIZE` and `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` to determine the size of a buffer to export a key pair or a public key respectively. Occasionally, it would be useful to have a constant for when you know you're working with a key pair or a public key, but you don't know which. Which one is bigger depends on what key types are supported.

This has come up in Mbed TLS [when writing code to bridge the PSA API with our legacy API for public-key cryptography](https://github.com/Mbed-TLS/mbedtls/pull/8774#discussion_r1474330273). It's not a common requirement, but it could come up when writing generic key management code that knows it's working with asymmetric cryptography, but manipulates both public-key and key-pair objects. Also there's a risk that someone needing this and used to RSA would assume that key pairs are larger, but they're actually not in an ECC-only configuration — or conversely someone used to ECC would make the opposite assumption. Since this is only a macro definition and will not affect an implementation's code size, we might as well include it in the official API.

Tentative name: `PSA_EXPORT_KEY_PAIR_OR_PUBLIC_KEY_MAX_SIZE` or `PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE`.

athoelke:
This simple change is currently based on the pending integration of the PAKE extension into version 1.3 (#177). The commit to review is https://github.com/ARM-software/psa-api/commit/545021be06af5fd90ba9a9755f172361d41ca98d.

Fixes #180

### DIFF
--- a/doc/crypto/api.db/psa/crypto.h
+++ b/doc/crypto/api.db/psa/crypto.h
@@ -219,6 +219,7 @@ typedef uint8_t psa_pake_step_t;
 #define PSA_ECC_FAMILY_TWISTED_EDWARDS ((psa_ecc_family_t) 0x42)
 #define PSA_ERROR_INSUFFICIENT_ENTROPY ((psa_status_t)-148)
 #define PSA_ERROR_INVALID_PADDING ((psa_status_t)-150)
+#define PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE /* implementation-defined value */
 #define PSA_EXPORT_KEY_OUTPUT_SIZE(key_type, key_bits) \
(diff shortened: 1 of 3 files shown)


MarcusJGStreets (approved):
Looks OK to me

athoelke:
Rebased after merging #177

PR #197, opened 2024-04-08
Responce to Andrew T's comments - still WIP

MarcusJGStreets:
Work in progress working through Andrew's comments.  need to add more details on what can be done during iteration.


MarcusJGStreets:
This was just a placeholder in case my synch did not work, use #148

PR #199, opened 2024-06-20
Interruptible APIs for key agreement use cases

### LINKED ISSUE #198: Support for interruptible key agreement/exchange
Issue #23 considers interruptible API design, and specifically looks at an API for asymmetric signature. There is a PR for an interruptible API for asymmetric signature creation and verification in #107.

Another use case that requires an interruptible operation is key agreement, in particular ECDH. (See https://github.com/Mbed-TLS/mbedtls/issues/9044)

For ECDH, the complex computation occurs at the following points:

1. Computing the public key from an ECC private key.
2. Validation of the public key provided by the other participant.
3. Computing the shared secret from the two input keys.

In the current PSA Crypto API, steps (2) and (3) occur as part of the key agreement function. Step (1) might occur during key generation, or during public key export, depending on the implementation's approach to storing ECC keys.

For full coverage of the key agreement use case we would need interruptible operations for all three APIs (key generation, key export, and key agreement).

As each of these functions have simple (bounded) inputs, the interruptible operations can be much simpler that the one designed for asymmetric signature. In particular, each operation only requires a single setup function taking all the inputs (called once) and a single completion function providing the outputs (called until finished).

athoelke:
Add interruptible operations for `psa_generate_key()`, `psa_export_public_key()` and `psa_key_agreement()`.

This provides the new API elements, and updates the *Functionality* chapter to cover this simpler interruptible pattern, as well as the complex interruptible multi-part operations used for signature and verification from #107.

This is based on top of #107. The first unique commit is https://github.com/ARM-software/psa-api/commit/57ae95c9dd6cd5aed55806d1aca3c021fc5dd8e7.

A preview of the PDF rendering of this PR: [IHI0086-PSA_Certified_Crypto_API-1.3.0-interruptible-operations-draft3.pdf](https://github.com/user-attachments/files/15927984/IHI0086-PSA_Certified_Crypto_API-1.3.0-interruptible-operations-draft3.pdf)


MarcusJGStreets (approved):
Look good to me

doc/crypto/api/ops/key-agreement.rst
```diff
@@ -332,6 +338,264 @@ Combining key agreement and key derivation
...
+        The peer key data is parsed with the type :code:`PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(type)` where ``type`` is the type of ``private_key``, and with the same bit-size as ``private_key``.
+        The peer key must be in the format that `psa_import_key()` accepts for this public key type.
+        These formats are described with the public key type in :secref:`key-types`.
+    .. param:: size_t peer_key_length
+        Size of ``peer_key`` in bytes.
+    .. param:: psa_algorithm_t alg
+        The standalone key agreement algorithm to compute: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_STANDALONE_KEY_AGREEMENT(alg)` is true.
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the new key.
+        This function uses the attributes as follows:
```
gilles-peskine-arm, line 448: The bullet points below actually mostly explain how the resulting key is constructed rather than how the `attributes` parameter is used. I find this presentation confusing. In particular, the key size is always the same regardless of the attributes. Separately, the key size in the attributes must be either correct or 0.

doc/crypto/api/ops/key-agreement.rst
```diff
@@ -332,6 +338,264 @@ Combining key agreement and key derivation
...
+            Implementations must support the `PSA_KEY_TYPE_DERIVE` and `PSA_KEY_TYPE_RAW_DATA` key types.
+
+        *   The size of the returned key is always the bit-size of the shared secret, rounded up to a whole number of bytes.
+            The key size in ``attributes`` can be zero; if it is nonzero, it must be equal to the output size of the key agreement, in bits.
+
+            The output size, in bits, of the key agreement is :code:`8 * PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE(type, bits)`, where ``type`` and ``bits`` are the type and bit-size of ``private_key``.
+
+        *   The key permitted-algorithm policy is required for keys that will be used for a cryptographic operation, see :secref:`permitted-algorithms`.
+        *   The key usage flags define what operations are permitted with the key, see :secref:`key-usage-flags`.
+        *   The key lifetime and identifier are required for a persistent key.
```
gilles-peskine-arm, line 461: The lifetime and identifier are also required for a volatile key, they just use different sets of values.

athoelke: That text is taken from _all_ the other key creation functions, and was first used when we clarified the use of attributes 4 years ago. (Private link to original source commit: https://github.com/ARM-software/psa-crypto-api/commit/5d12b9abdcb1cc183a0ba8113771a034ecc6fd64).

Perhaps it would be clearer to state that the lifetime and identifier have to be set explicitly for a persistent key, as the default values will create a volatile key?

doc/crypto/api/ops/key-agreement.rst
athoelke: This text is identical to that for `psa_key_agreement()`, which we discussed at some length as part of https://github.com/ARM-software/psa-api/pull/101 (specifically commit https://github.com/ARM-software/psa-api/pull/101/commits/5c48afe702061b077754cf60156b4b1972713742).

Is there a good reason to do something different here, or to change what we have done in the existing APIs?

gilles-peskine-arm: Sorry, I missed that earlier. My objection isn't specific to this function. The explanations of how the attributes are used and how the new key is constructed are confusing.

athoelke: Proposal for improvement raised as #205. Will feed results of that PR into this one.

doc/crypto/api/ops/key-agreement.rst
athoelke: Proposal for improvement raised as #205

doc/crypto/api/ops/key-agreement.rst
athoelke: Clarifications in the updated PR #205 have been copied into this PR

doc/crypto/api/keys/management.rst
```diff
@@ -529,3 +536,456 @@ Key export
...
+
+    .. param:: psa_generate_key_iop_t * operation
+        The interruptible key generation operation to set up.
+        It must have been initialized as per the documentation for `psa_generate_key_iop_t`, and be inactive.
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the new key.
+
+        The following attributes are required for all keys:
+
+        *   The key type. It must not be an asymmetric public key.
```
gilles-peskine-arm, line 638: This suggests that implementations that support interruptible operations should support interruptible key generation for symmetric keys, even though operations _on_ symmetric keys don't have interruptible APIs. I think we should say that only key-pair types are expected to be supported, and implementations are expected to reject symmetric keys but they may accept them.

gilles-peskine-arm: Although on second thoughts, it's probably one less line of code to just systematically fall back to `psa_generate_key` for any key type that doesn't have interruptible support, and let it decide between public keys, key pairs with single-iop generation, and symmetric keys.

athoelke: It might be useful to state that an implementation is permitted to only support the use of this API for a subset of the key types that can be generated using `psa_generate_key()`?

athoelke: I've added a commit to note that implementations might only provide incremental generation for some key types.

athoelke:
Rebased on main

doc/crypto/api/ops/key-agreement.rst
```diff
@@ -350,6 +356,276 @@ Combining key agreement and key derivation
...
+        Get the number of *ops* that an interruptible key agreement operation has taken so far.
+
+    .. param:: psa_key_agreement_iop_t * operation
+        The interruptible key agreement operation to inspect.
+
+    .. return:: uint32_t
+        Number of *ops* that the operation has taken so far.
+
+    After the interruptible operation has completed, the returned value is the number of *ops* required for the entire operation.
+    The value is reset to zero by a call to either `psa_key_agreement_iop_setup()` or `psa_key_agreement_iop_abort()`.
```
gilles-peskine-arm, line 438: When we were discussing the specification, we implicitly assumed that setup takes negligible time. But this is not a safe assumption. For example, for key agreement, the public key needs to be validated. In Mbed TLS, we do this during setup (and it's the natural way to do it). Mbed TLS estimates the public key validation to take 3 ops. (The key agreement as a whole takes thousands of ops.)

So I think we should specify that setup resets the op count to 0 at the beginning of its operation, but then adds the number of ops that it takes. And recommend to implementations that this number should be lower than the value set with `psa_iop_set_max_ops`.

gilles-peskine-arm: This applies to all the functions, although key agreement might be the only one for which it's really natural to have setup take a few ops.

athoelke: That does complicate the way we can describe the reset of the _ops_ value, but is a valid point. Is it sufficient to expect that validation will occur during setup, or should the recommendation be that if validation can be very complex (substantially more _ops_ than a reasonable application-configured maximum) it should be deferred to the completion function?

Similar to the case for deferring key id allocation (`PSA_ERROR_ALREADY_EXISTS` returned by `psa_key_agreement_iop_complete()`), should we also permit `psa_key_agreement_iop_complete()` to return status codes related to the validation of the peer key? - which enables deferred validation.

gilles-peskine-arm: I think we should recommend that setup should only perform a small number of ops, preferably less than whatever was set by `psa_iop_set_max_ops()`.

And I hadn't thought of that, but indeed we should allow complete() to return status codes related to validations that may be considered slow.

athoelke:
Rebased, and applied documentation improvements that have been merged to `main`.

athoelke:
Addressed the comment (https://github.com/ARM-software/psa-api/pull/107#discussion_r1882310284) in #107, as this PR has an updated version of that chapter, and the comment applies to additional parts of the text.

athoelke:
This PR has been superseded by #375, which combines the signature and additional nterruptible operations as a PR against the latest specification version.

PR #205, opened 2024-07-23
Further clarifications on the use of key attributes when creating keys

athoelke:
An attempt to improve the documentation of how to set the attributes when creating a key, and how the implementation must use them.

In particular:

* Removing the sloppy text relating to lifetimes and identifiers for key
* Making it clearer when attributes are not required to be set. For example, the key size for key import, key agreement, pake etc.

### DIFF
--- a/doc/crypto/api/keys/management.rst
+++ b/doc/crypto/api/keys/management.rst
@@ -36,20 +36,33 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
 
     .. param:: const psa_key_attributes_t * attributes
         The attributes for the new key.
-        This function uses the attributes as follows:
 
-        *   The key type is required, and determines how the ``data`` buffer is interpreted.
-        *   The key size is always determined from the ``data`` buffer. If the key size in ``attributes`` is nonzero, it must be equal to the size determined from ``data``.
-        *   The key permitted-algorithm policy is required for keys that will be used for a cryptographic operation, see :secref:`permitted-algorithms`.
-        *   The key usage flags define what operations are permitted with the key, see :secref:`key-usage-flags`.
-        *   The key lifetime and identifier are required for a persistent key.
+        The following attributes are required for all keys:
+
+        *   The key type determines how the ``data`` buffer is interpreted.
+
+        The following attributes must be set for keys used in cryptographic operations:
+
+        *   The key permitted-algorithm policy, see :secref:`permitted-algorithms`.
+        *   The key usage flags, see :secref:`key-usage-flags`.
+
+        The following attributes must be set for keys that do not use the default volatile lifetime:
+
+        *   The key lifetime, see :secref:`key-lifetimes`.
+        *   The key identifier is required for a key with a persistent lifetime, see :secref:`key-identifiers`.
+
+        The following attributes are optional:
+
+        *   If the key size is nonzero, it must be equal to the key size determined from ``data``.
 
         .. note::
-            This is an input parameter: it is not updated with the final key attributes. The final attributes of the new key can be queried by calling `psa_get_key_attributes()` with the key's identifier.
+            This is an input parameter: it is not updated with the final key attributes.
+            The final attributes of the new key can be queried by calling `psa_get_key_attributes()` with the key's identifier.
 
     .. param:: const uint8_t * data
         Buffer containing the key data.
         The content of this buffer is interpreted according to the type declared in ``attributes``.
+
         All implementations must support at least the format described in the *Key format* section of the chosen key type.
         Implementations can support other formats, but be conservative in interpreting the key data: it is recommended that implementations reject content if it might be erroneous, for example, if it is the wrong type or is truncated.
     .. param:: size_t data_length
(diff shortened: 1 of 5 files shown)


athoelke:
Once these changes are agreed as an improvement, I will propagate the changes to the new key creation functions in #194 and #199

doc/crypto/api/keys/management.rst
```diff
@@ -216,7 +254,7 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
     .. retval:: PSA_ERROR_BAD_STATE
         The library requires initializing by a call to `psa_crypto_init()`.
 
-    Copy key material from one location to another.
+    Copy key material from one location to another. Its location is taken from ``attributes``, its policy is determined by both ``attributes`` and the source key policy, and its type and size are taken from the source key.
```
gilles-peskine-arm, line 257: The description of the resulting key policy is vague and could be read as e.g. the union of the two policies rather than the intersection. The policy is _constrained_ by both `attributes` and the source key policy, although that may not be a comprehensible way to put it. In plain language, the resulting key can do everything that is both permitted for the original key and permitted by `attributes`.

doc/crypto/api/ops/key-derivation.rst
```diff
@@ -847,19 +847,27 @@ Key derivation functions
...
+        The following attributes are required for all keys:
+
+        *   The key type. It must not be an asymmetric public key.
+        *   The key size. It must be a valid size for the key type.
+
+        The following attributes must be set for keys used in cryptographic operations:
+
+        *   The key permitted-algorithm policy, see :secref:`permitted-algorithms`.
 
             If the key type to be created is `PSA_KEY_TYPE_PASSWORD_HASH`, then the permitted-algorithm policy must be the same as the current operation's algorithm.
```
gilles-peskine-arm, line 860: Preexisting: technically, a more restrictive policy would also be ok. We currently have no wildcards that would be applicable here, but 0 should be allowed (even if I don't see a concrete use for it).

doc/crypto/api/ops/pake.rst
```diff
@@ -954,23 +954,30 @@ Multi-part PAKE operations
...
 
-        *   The key type is required.
+        The following attributes are required for all keys:
+
+        *   The key type.
             All PAKE algorithms can output a key of type :code:`PSA_KEY_TYPE_DERIVE` or :code:`PSA_KEY_TYPE_HMAC`.
             PAKE algorithms that produce a pseudo-random shared secret, can also output block-cipher key types, for example :code:`PSA_KEY_TYPE_AES`.
             Refer to the documentation of individual PAKE algorithms for more information.
-        *   The key size in ``attributes`` must be zero.
+        *   The key size in ``attributes`` must be zero --- the default value.
```
gilles-peskine-arm, line 964: Preexisting: the correct value should be allowed.

The intent is that if you obtain a key's attributes with `psa_get_key_attributes`, what you obtain should always be valid to create a similar key.

gilles-peskine-arm (changes requested):
Looks mostly good to me

doc/crypto/api/ops/pake.rst
athoelke: In this case, I think I recall that we didn't want to try and construct a helper macro that could report the actual key size that would be returned.

But perhaps the absence of a macro to compute a value from a provided cipher-suite, is not a good enough reason to refuse to accept a correct value. A caller who knows precisely what cipher suite they are using could specify the known key size here - and it would be counter-intuitive if that failed.

So I think I agree with relaxing this.

doc/crypto/api/keys/management.rst
athoelke: This first paragraph is aiming to be a summary of how the attributes are derived (cf the other key creation functions), and the full detail regarding policies is provided a few paras later in this description.

I'd rather not bloat this sentence, but would be happy to find a short phrase that more accurately describes the combination/intersection of policies for use here, that reduces the misunderstanding of the process.

doc/crypto/api/ops/key-derivation.rst
athoelke: The preceding text "for keys used in a cryptographic operation" would eliminate the zero case - as all operation functions demand one usage or another to be permitted. I think that the only thing one can do with a key with no usage flags is to export the public key, or destroy it (I haven't checked that, so might be missing something). Not sure it is worth documenting a technically valid, but entirely useless scenario for the specification?

I am inclined to leave this as is, and worry about relaxing the constraint when we have a relevant wildcard policy that would make the more complex-to-describe behavior valuable.

athoelke (commented):
_Ooops: comment added to wrong place_

doc/crypto/api/ops/key-derivation.rst
athoelke: I note now that this is not about `0` usage flags, but for a default no-permitted-algorithm policy. Keys like this can still be exported (if they have `PSA_KEY_USAGE_EXPORT`), and so can act as a secure storage for secrets.

There is also a test use case for simplified testing of key creation functionality where the construction of the key is tested by examining the resulting value, but the test suite can simply request no specific permitted algorithm.

I could reword this as a double negative, which perhaps implies that a 'no algorithm' policy is valid (but is not explicit):

> If the key type to be created is `PSA_KEY_TYPE_PASSWORD_HASH`, then the attributes must not specify a permitted-algorithm policy that is not the algorithm used to compute the password hash.

But that does not seem clear. If we wish to allow 'no permitted algorithm', then I think it would be better to explicitly include it in a positive specification rule:

> If the key type to be created is `PSA_KEY_TYPE_PASSWORD_HASH`, then the permitted-algorithm policy must either be the same as the current operation's algorithm, or `PSA_ALG_NONE`.

Is " the algorithm used to compute the password hash" as clear as "the current operation's algorithm"? - the former provides the rationale for the rule, but the implementor has to deduce that this is the operation algorithm; the latter is precise but the justification requires a deductive step.

doc/crypto/api/ops/pake.rst
athoelke: I have now permitted the correct size

gilles-peskine-arm (approved):
LGTM

PR #206, opened 2024-07-23
Clarified flexibility for unsupported parameter values

athoelke:
Provide additional information about the flexibility in the Crypto API in relation to the use of `PSA_ERROR_INVALID_ARGUMENT` and `PSA_ERROR_NOT_SUPPORTED`.

### DIFF
--- a/doc/crypto/api/library/status.rst
+++ b/doc/crypto/api/library/status.rst
@@ -45,7 +45,7 @@ These definitions must be available to an application that includes the :file:`p
 Common error codes
 ^^^^^^^^^^^^^^^^^^
 
-Some of the common status codes have a more precise meaning when returned by a function in the |API|, compared to the definitions in `[PSA-STAT]`.
+Some of the common status codes have a more precise meaning when returned by a function in the |API|, compared to the definitions in `[PSA-STAT]`. See also :secref:`error-handling`.
 
 .. list-table::
     :class: longtable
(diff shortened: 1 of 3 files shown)


Vge0rge:
That's a very good improvement! it's very useful and it simplifies the drivers to allow returning the error code that they see fit, thank you Andrew!

athoelke:
> simplifies the drivers to allow returning the error code that they see fit

I think that the behavior for implementations of the Crypto Driver API needs to be more tightly controlled. In particular, a driver entry-point should return `PSA_ERROR_NOT_SUPPORTED` for any algorithm, key type, or combination that it does not support or does not recognize.

This is because an implementation of the application API can include multiple drivers, and it makes the design of the core/dispatching code better if a `PSA_ERROR_INVALID_ARGUMENT` status code clearly indicates that the call should fail, but `PSA_ERROR_NOT_SUPPORTED` indicates that the dispatcher should check if another driver can handle the request.

Vge0rge:
> I think that the behavior for implementations of the Crypto Driver API needs to be more tightly controlled. In particular, a driver entry-point should return PSA_ERROR_NOT_SUPPORTED for any algorithm, key type, or combination that it does not support or does not recognize.

I agree with this, a driver should check only for valid/invalid configurations which it supports.

doc/crypto/api/library/status.rst
```diff
@@ -55,6 +55,16 @@ Some of the common status codes have a more precise meaning when returned by a f
...
 
+    * - :code:`PSA_ERROR_NOT_SUPPORTED`
+      - `[PSA-STAT]` recommends the use of :code:`PSA_ERROR_INVALID_ARGUMENT` for invalid parameter values.
+
+        In the |API|, this is relaxed for algorithm identifier and key parameters, when either :code:`PSA_ERROR_INVALID_ARGUMENT` or :code:`PSA_ERROR_NOT_SUPPORTED` can be returned.
+
+    * - :code:`PSA_ERROR_INVALID_ARGUMENT`
+      - `[PSA-STAT]` recommends the use of :code:`PSA_ERROR_NOT_SUPPORTED` for unsupported parameter values.
+
+        In the |API|, either :code:`PSA_ERROR_INVALID_ARGUMENT` or :code:`PSA_ERROR_NOT_SUPPORTED` can be returned when unsupported algorithm identifier or key parameters are used.
```
gilles-peskine-arm, line 66: As above:

```suggestion
        In the |API|, either :code:`PSA_ERROR_INVALID_ARGUMENT` or :code:`PSA_ERROR_NOT_SUPPORTED` can be returned when unsupported algorithm identifier or key type parameters are used. This allows implementations to avoid having to recognize all the cryptographic mechanisms that are defined in the PSA specification but not provided by that particular implementation.
```

gilles-peskine-arm (changes requested):
Missing word, and an explanation.

athoelke:
@gilles-peskine-arm - thanks for the review, I have taken your suggestions.

gilles-peskine-arm (approved):
LGTM

PR #213, opened 2024-09-04
Clarified the constraints on key derivation output and verification functions

### LINKED ISSUE #212: Clarify constraints on key output from a key derivation operation
The documentation for [`psa_key_derivation_output_key()`](https://arm-software.github.io/psa-api/crypto/1.2/api/ops/kdf.html#c.psa_key_derivation_output_key), states that this is not permitted if a `PSA_KEY_DERIVATION_INPUT_SECRET` step did not use `psa_key_derivation_input_key()` or `psa_key_derivation_key_agreement()`. This is both in the error condition table, and the description.

This constraint is also true for `PSA_KEY_DERIVATION_INPUT_PASSWORD` input steps - see [the description of this step](https://arm-software.github.io/psa-api/crypto/1.2/api/ops/kdf.html#c.PSA_KEY_DERIVATION_INPUT_PASSWORD). However, this is not documented in the `psa_key_derivation_output_key()` function, with the `PSA_KEY_DERIVATION_INPUT_SECRET` constraint, only in the definition of `PSA_KEY_DERIVATION_INPUT_PASSWORD`.

This omitted constraint description is also missing from the `psa_key_derivation_output_key_custom()` API introduced in #194.

athoelke:
Clarifications in the key derivation operation:

* Clarified the constraint on `psa_key_derivation_output_key()` for algorithms that have a `PSA_KEY_DERIVATION_INPUT_PASSWORD` input step.
* Removed the redundant key input constraints on `psa_key_derivation_verify_bytes()` and `psa_key_derivation_verify_key()`. These match the policy already checked in `psa_key_derivation_input_key()`.

Fixes #212

### DIFF
--- a/doc/crypto/api/ops/key-derivation.rst
+++ b/doc/crypto/api/ops/key-derivation.rst
@@ -758,7 +758,10 @@ Key derivation functions
     .. retval:: PSA_ERROR_INVALID_HANDLE
         ``key`` is not a valid key identifier.
     .. retval:: PSA_ERROR_NOT_PERMITTED
-        The key has neither the `PSA_KEY_USAGE_DERIVE` nor the `PSA_KEY_USAGE_VERIFY_DERIVATION` usage flag, or it does not permit the operation's algorithm.
+        The following conditions can result in this error:
+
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm (approved):
LGTM

PR #214, opened 2024-09-06
Documented use of context parameters in PAKE operations

### LINKED ISSUE #209: Missing information about contexts in PAKE protocols
Some PAKE algorithms use a context parameter, and there is an API for each participant to set the context value. However, none of the documented PAKE protocols describe if, or how this parameter should be set, or what happens if none is set.

### J-PAKE

J-PAKE does not use a context. The documentation should state that setting the context is an error for J-PAKE.

### SPAKE2+

SPAKE2+ does have a context, which is used when generated the transcript input to the key derivation process.

- For the MATTER PAKE the context is fully defined in the Matter specification, and the caller is required to construct the Context string as specified there, and pass it to the PAKE operation.

- The general SPAKE2+ algorithm RFC that the context is an octet string that can be `nil` (the empty string), and may alternatively be omitted (along with the preceding encoded `len(context)`) from the transcript.

We presently say nothing in the specification, other than showing `psa_pake_set_context()` being used in the SPAKE2+ protocol flow.

So I'd say that a clarification is necessary for both the MATTER and RFC variants of SPAKE2+. For the latter, we need to choose between the following ways of handling omitted and zero-length context values:

1. A zero-length context is treated the same as not setting a context: both are encoded as `len(nil) || nil || ...` in the transcript.
2. A zero-length context is treated the same as not setting a context: both are omitted from the transcript.
3. If no context is set, the context and encoded-length are omitted from the transcript; if a zero-length context is set, it is encoded as `len(nil) || nil || ...` in the transcript.

Option 3 gives the full range of possible transcript encodings to the application; though it creates a different behavior to the handling of user-id and peer-id which treat an unset value as a zero-length string.

Have any implementations made a specific choice for this parameter to SPAKE2+?

athoelke:
Following option 3 in #209.

Fixes #209

### DIFF
--- a/doc/crypto/api/ops/pake.rst
+++ b/doc/crypto/api/ops/pake.rst
@@ -1233,7 +1233,7 @@ J-PAKE requires both an application and a peer identity.
 If the peer identity provided to `psa_pake_set_peer()` does not match the data received from the peer, then the call to `psa_pake_input()` for the `PSA_PAKE_STEP_ZK_PROOF` step will fail with :code:`PSA_ERROR_INVALID_SIGNATURE`.
 
 J-PAKE does not use a context.
-A call to `psa_pake_set_context()` for a J-PAKE operation will fail with :code:`PSA_ERROR_INVALID_ARGUMENT`.
+A call to `psa_pake_set_context()` for a J-PAKE operation will fail with :code:`PSA_ERROR_BAD_STATE`.
 
 The following steps demonstrate the application code for 'User' in :numref:`fig-jpake`. The code flow for the 'Peer' is the same as for 'User', as J-PAKE is a balanced PAKE.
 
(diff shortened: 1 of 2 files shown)


athoelke:
Feedback appreciated from @silabs-Kusumit, @silabs-hannes, @oberon-sk, and @yanesca.

doc/crypto/api/ops/pake.rst
```diff
@@ -1230,6 +1232,9 @@ J-PAKE does not assign roles to the participants, so it is not necessary to call
 J-PAKE requires both an application and a peer identity.
 If the peer identity provided to `psa_pake_set_peer()` does not match the data received from the peer, then the call to `psa_pake_input()` for the `PSA_PAKE_STEP_ZK_PROOF` step will fail with :code:`PSA_ERROR_INVALID_SIGNATURE`.
 
+J-PAKE does not use a context.
+A call to `psa_pake_set_context()` for a J-PAKE operation will fail with :code:`PSA_ERROR_INVALID_ARGUMENT`.
+
```
oberon-sk, line 1237: I would prefer `PSA_ERROR_BAD_STATE` in this case. This is analog to `psa_cipher_set_iv()`, where `PSA_ERROR_BAD_STATE` shall be returned if the _cipher algorithm does not use an IV_.

athoelke: I think I agree. We should also change the text for `psa_pake_set_context()` - which currently describes `PSA_ERROR_INVALID_ARGUMENT` as being returned if the protocol does not support a context.

I do not think that the incompatibility (changing the specific error code) is an issue in practice here.

PR #215, opened 2024-09-10
DRAFT: Support for key wrapping

### LINKED ISSUE #50: Add APIs to support wrapped keys
Many secure elements and crypto accelerators require the use of wrapped keys and will not accept importing clear-text keys to their key store. The existing `psa_import_key` function could be augmented to support wrapped keys, but that puts the burden of identifying wrapping and associated parameters on the underlying implementation. A new function should be proposed to support wrapped key imports.
As there is no standard for key wrapping data formats and associated algorithms, this should be made generic enough to be adaptable to any such key stores.
Same need for exporting wrapped keys.

athoelke:
_[Added 24/10/24]_ Superseded by #224 - which provides just the simplified key wrapping and unwrapping functionality. The wrapped-key format support will be represented in a later PR.

----

Building on the key formats in #207, this PR defines an API for wrapping and unwrapping keys.

The first attempt explores an approach which extends the `psa_import_formatted_key()` and `psa_export_formatted_key()` APIs, and adds a key wrapping key and algorithm to the process. The idea is that it could support wrapping keys using a general key-wrapping algorithm (such as AES-KW) with existing clear-text key formats, and wrapping keys in wrapped-key data formats, such as *EncryptedPrivateKeyInfo*.

I am publishing this as an initial draft to provoke discussion around the challenges and undecided issues with this approach, including the policy integration issue that affects import of formatted keys.

The first commit specific to this PR (not in #207) is https://github.com/ARM-software/psa-api/pull/215/commits/192eb55088b13b4b3bda07eff978762849b44d61

Fixes #50.

### DIFF
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -250,6 +250,28 @@ The usage flags are encoded in a bitmask, which has the type `psa_key_usage_t`. 
 
     If this flag is present on all keys used in calls to `psa_key_derivation_input_key()` for a key derivation operation, then it permits calling `psa_key_derivation_verify_bytes()` or `psa_key_derivation_verify_key()` at the end of the operation.
 
+.. macro:: PSA_KEY_USAGE_WRAP
+    :definition: ((psa_key_usage_t)0x00010000)
+
+    .. summary::
+        Permission to wrap another key with the key.
+
+    This flag is required to use the key in a key-wrapping operation.
+    The flag must be present on keys used with the following APIs:
+
+    *   `psa_wrap_key()`
+
+.. macro:: PSA_KEY_USAGE_UNWRAP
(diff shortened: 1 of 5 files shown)


athoelke:
I think we need to identify a representative set of use cases to cover the important ways in which the API needs to support key wrapping. And ensure we consider agility for future key types (PAKE, PQC, etc.) as part of the process.

It may be that we need more than one API to handle standard wrapped-key formats that are self-describing (in terms of wrapping algorithm(s)); general purpose key wrapping algorithms (typically specialized AEAD algorithms); and applications that use AEAD algorithms to create non-standard key-wrapping schemes.

athoelke:
I'm thinking about a change of plan here. See [my comment](https://github.com/ARM-software/psa-api/issues/50#issuecomment-2401779444) in PR #207.

PR #218, opened 2024-09-23
FIPS 204 - API for Module Lattice-based signatures (ML-DSA)

### LINKED ISSUE #96: Define API for Module-Lattice-Based Digital Signature (ML-DSA aka CRYSTALS-Dilithium)
**Updated 13-09-2024:** The final version of the specification is now published https://csrc.nist.gov/pubs/fips/204/final.

NIST has now published a draft specification for a Digital Signature algorithm derived from CRYSTALS-Dilithium. The algorithm is designated ML-DSA, and will be published as FIPS 204. The draft (open for review until 22 Nov 2023) can be downloaded from https://csrc.nist.gov/pubs/fips/204/ipd.

NIST is planning to standardize three parameterized variants of ML-DSA: ML-DSA-44, ML-DSA-65, ML-DSA-87, which provide increasing levels of security for increasing computation and size of keys and signature values.

The signature algorithm is recommended to use fresh randomness, to mitigate side-channel attacks; but can also operate deterministically when the implementation does not have access to randomness, by substituting a zero value in place of the random value.

The Crypto API should define an API for using these algorithms.

athoelke:
Based on the discussion and proposal in #96 and #210, here is a definition of key types and algorithms for ML-DSA.

I have extracted the ML-DSA definitions out of the #210 PR, in which they are entangled with ML-KEM and ECIES API definitions, in order to fill out the remaining API for ML-DSA, using a similar structure to the SLH-DSA definitions in #216.

* The three parameter sets are supported
* Four algorithm variants for Hedged/Deterministic and Pure/Pre-hash options. The provision of explicit deterministic variants is an active discussion topic, see #210 and #216.
* Context values are not supported. These need an additional API to provide the context - as per existing sign-with-optional-context algorithms.

Fixes #96

### DIFF
--- a/doc/crypto/api/keys/types.rst
+++ b/doc/crypto/api/keys/types.rst
@@ -1165,19 +1165,12 @@ The curve type affects the key format, the key derivation procedure, and the alg
     .. return:: psa_ecc_family_t
         The elliptic curve family id, if ``type`` is a supported elliptic curve key. Unspecified if ``type`` is not a supported elliptic curve key.
 
-.. _ml-keys:
+.. _ml-dsa-keys:
 
-Module Lattice keys
--------------------
+Module Lattice-based Signature keys
+-----------------------------------
 
-The |API| supports Module Lattice cryptography as defined in :cite-title:`FIPS203` and :cite-title:`FIPS204`.
-
-There are two related, but separate, algorithms:
-
-*   A key encapsulation method, ML-KEM.
-*   A signature method, ML-DSA.
-
-These algorithms have distinct asymmetric key types.
+The |API| supports Module Lattice-based digital signatures (ML-DSA), as defined in :cite-title:`FIPS204`.
 
 .. macro:: PSA_KEY_TYPE_ML_DSA_KEY_PAIR
     :definition: ((psa_key_type_t)0x7002)
--- a/doc/crypto/api/keys/types.rst
+++ b/doc/crypto/api/keys/types.rst
@@ -1198,7 +1198,14 @@ The |API| supports Module Lattice-based digital signatures (ML-DSA), as defined 
 
     .. subsection:: Key format
 
-        An ML-DSA key pair is the :math:`(pk,sk)` pair of public key and secret key, which are generated from a secret 32-byte seed, :math:`\xi`. See [FIPS204]` §5.1.
+        .. warning::
+
+            The key format may change in a final version of this API.
+            The standardization of exchange formats for ML-DSA public and private keys is in progress, but final documents have not been published.
+
+            The current proposed format is based on the current expected outcome of that process.
+
+        An ML-DSA key pair is the :math:`(pk,sk)` pair of public key and secret key, which are generated from a secret 32-byte seed, :math:`\xi`. See `[FIPS204]` §5.1.
 
         The data format for import and export of the key pair is the 32-byte seed :math:`\xi`.
 
@@ -1221,8 +1228,6 @@ The |API| supports Module Lattice-based digital signatures (ML-DSA), as defined 
                     ML-DSA-87, 7488
 
             *   It is better for the standard to choose a single format to improve interoperability.
-
-            One consequence is that to enable future export of the private key (for example, a wrapped copy in a CMS or *EncryptedPrivateKeyInfo* structure), the implementation MUST store the seed :math:`\xi`, even if it also computes and stores the :math:`(pk,sk)` pair.
 
         See `PSA_KEY_TYPE_ML_DSA_PUBLIC_KEY` for the data format used when exporting the public key with `psa_export_public_key()`.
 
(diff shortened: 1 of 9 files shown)


doc/crypto/api/ops/signature.rst
```diff
@@ -347,6 +347,248 @@ Asymmetric signature algorithms
...
+`[FIPS204]` defines pure and pre-hashed variants of the signature scheme, which can either be hedged (randomized) or deterministic.
+Four algorithms are defined to support these variants: `PSA_ALG_ML_DSA`, `PSA_ALG_DETERMINISTIC_ML_DSA`, `PSA_ALG_HASH_ML_DSA()`, and `PSA_ALG_DETERMINISTIC_HASH_ML_DSA()`.
+
+*   The pre-hashed signature computation generates distinct signatures to a pure signature, even with the same key and message hashing algorithm.
+
+    When verifying a signature it is necessary to know which algorithm was used to generate it.
+
+*   Hedging incorporates fresh randomness in the signature computation, resulting in distinct signatures on every signing operation when given identical inputs.
+    Deterministic signatures do not require additional random data, and result in an identical signature for the same inputs.
+
```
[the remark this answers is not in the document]
athoelke: Please review the updated text. I still need to add some guidance regarding recommended algorithms - although that should really point to the standards, or will be decided the higher level protocol that is using ML-DSA.

doc/crypto/api/ops/signature.rst
```diff
@@ -347,6 +347,276 @@ Asymmetric signature algorithms
...
+.. todo::
+    Decide if these general comments are better kept in a common place (here), or if they should be repeated with as appropriate in each of the four algorithm definitions?
+
+.. rationale::
+
+    The use of fresh randomness, or not, when computing a signature looks primarily like an implementation decision based on the capability of the system, and its vulnerability to specific threats, following the recommendations in `[FIPS204]`.
+
+    However, the |API| gives distinct algorithm identifiers for the hedged and deterministic variants of both ML-DSA and HashML-DSA, for the following reasons:
+
+    *   `[FIPS204]` recommends that ML-DSA signing keys are only used to compute either deterministic, or hedged, signatures, but not both.
```
MarcusJGStreets, line 397: FIPS 205 includes this for SHS-DSA 
FIPS 204 does not. 
It simply says "this variant _should not_ be used on platforms where side-channel attacks are a concern"

The History notes that Dilithium was originally only deterministic. 
The hedged version was added as a defence against side-channel attacks.

athoelke: Technically, the full quote is:

>Therefore, this variant **should not** be used on platforms where side-channel attacks are a concern and where they cannot be otherwise mitigated.

There is an extra clause that modifies the applicability of the recommendation to contexts where this threat is relevant (confidentiality of messages is an application issue), and not otherwise mitigated (a platform/system issue).

athoelke: It seems that the basic guidance for applications is "prefer to use hedged, because this is the default in the standard", but "if the implementation does not support it, AND you can accept the residual risk to message confidentiality, then use deterministic instead".

The guidance for implementations is "Always provide hedged, unless you have absolutely no viable source of randomness", and "provide deterministic if hedged cannot be provided, or as an extra optional variant alongside hedged".

MarcusJGStreets: It does say "even a weak RBG may be preferable to the fully deterministic variants"

How many implementations would not even have a weak RNG?
In that case, how much other crypto can they do?

doc/crypto/api/keys/types.rst
```diff
@@ -1165,6 +1165,122 @@ The curve type affects the key format, the key derivation procedure, and the alg
...
+
+    .. subsection:: Compatible algorithms
+
+        .. hlist::
+
+            *   `PSA_ALG_ML_DSA`
+            *   `PSA_ALG_HASH_ML_DSA`
+            *   `PSA_ALG_DETERMINISTIC_ML_DSA`
+            *   `PSA_ALG_DETERMINISTIC_HASH_ML_DSA`
+
```
MarcusJGStreets, line 1198: I would like to add a note on hedged versus deterministic here. Perhaps:

.. admonition:: Note 

   FIPS 204 defines the "hedged" and deterministic versions. The hedged version provides message secrecy and some protection against side-channels. FIPS 204 recommends that users *should* use the hedged version if either of these issues are a concern. The deterministic variant *should* only be used if the implementation does not include any source of randomness. 

.. admonition:: Implementation note

   FIPS 204 recommends that implementations use an approved random number generator to provide the random value in the hedged version. However, it notes that use of the hedged variant with a weak RNG is generally preferable to the deterministic variant.

athoelke: I'm not sure that putting it here, with the key types is the right place - these should be with the algorithms. Perhaps this text would work well in the ML-DSA algorithm overview where it discusses Hedged and Deterministic variants?

I would leave the warning (from the last commit) in the description of `PSA_ALG_DETERMINISTIC_ML_DSA`.

athoelke:
I've reworked the algorithm intro, including the suggested notes for application and implementation devs.

MarcusJGStreets:
Having looked art the last commit, I think that is reasonable.

athoelke:
The work of the IETF LAMPS WG to define how ML-DSA appears in PKIX infrastructure is in progress. The flexibility in FIPS204 for storing/exchanging private keys as a seed or a key-pair has resulted in a likely change of direction from the July draft towards using the 32-byte seed.

See https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/ for the latest status of this paper, and links to the drafts and the mailing list discussions.

We would prefer to NOT finalize the Crypto API definitions for key formats until this working group has published a final specification for key exchange formats.

This could result in publishing a PQC extension for the Crypto API, instead of integrating directly to the main specification. The extension will remain in Beta until the LAMPS WG work is complete, and perhaps also until we have confirmed the API through an implementation.

athoelke:
The replacement PR #223 has now been merged

PR #224, opened 2024-10-24
Add support for key wrapping (v2)

### LINKED ISSUE #50: Add APIs to support wrapped keys

athoelke:
This PR replaces #215, simplifying the API to only support key wrapping algorithms, and not wrapped-key formats. Wrapped-key formats will be considered later, along with import and export of formatted key data (see also #50, #149 and #207 for more discussion).

This added the `psa_wrap_key()` and `psa_unwrap_key()` functions, and support for AES-KW and AES-KWP algorithms.

Open issues:

1. NIST 800-38F describes a generalised AES-KW and AES-KWP to support any 128-bit 'approved block cipher', which exactly matches RFC 3394 and RFC 5649 definitions for the AES block cipher. Is it valuable to define this as the more general algorithm identifier, compatible with any 128-bit block cipher key, or just keep the AES-specific name and AES-only compatibility?

2. I have allocated a new algorithm category, as key-wrapping algorithms tend to be specialized authenticated encryption. However, AES-SIV can be used as a general AEAD algorithm, as well as a key-wrapping algorithm. Would that be problematic?

3. I have recycled the 'S' bit in the algorithm identifier to flag whether the algorithm has alignment constraints on the input data (AES-KW does, AES-KWP does not) - similar to block-aligned sizes for some block-mode ciphers. Is this at all useful, or should we just make this bit 0 for key-wrap algorithms?

Partly fixes #50


doc/crypto/api.db/psa/crypto.h
```diff
@@ -375,6 +380,9 @@ typedef struct psa_custom_key_parameters_t {
     /* implementation-defined value */
 #define PSA_TLS12_ECJPAKE_TO_PMS_OUTPUT_SIZE 32
 #define PSA_TLS12_PSK_TO_MS_PSK_MAX_SIZE /* implementation-defined value */
+#define PSA_WRAP_KEY_OUTPUT_SIZE(wrap_key_type, alg, key_type, key_bits) \
+    /* implementation-defined value */
+#define PSA_WRAP_KEY_PAIR_MAX_SIZE /* implementation-defined value */
```
oberon-sk, line 385: I would expect `PSA_WRAP_KEY_OUTPUT_MAX_SIZE` instead of `PSA_WRAP_KEY_PAIR_MAX_SIZE`.

athoelke: `psa_wrap_key()` has the same issue as `psa_export_key()` in relation to a MAX_SIZE macro for the output. There is no really helpful limit on the size of what can be stored in the keystore, for example certificates *could* be stored as `PSA_KEY_TYPE_RAW_DATA`. The description for the output buffer size states:

>* This API defines no maximum size for wrapped symmetric keys. Arbitrarily large data items can be stored in the key store, for example certificates that correspond to a stored private key or input material for key derivation.

For key export, we have `PSA_EXPORT_KEY_PAIR_MAX_SIZE`, `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` and (new for 1.3) `PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE` (which is just the maximum of the other two). The reason for separate key-pair and public-key sizes is that these can be very different values, and most use cases would know which of these is required, permitting a more efficient choice of buffer size.

For key wrapping, I didn't include a macro for wrapped public keys, or asymmetric keys in general as I expected that there is no real use case for simple wrapping of public keys. If this assumption is wrong, I would be happy to add the additional macros to the API, or we might determine that a single one for wrapped asymmetric key maximum size is sufficient.

oberon-sk: Agreed, there is no general maximum size for export and key_wrap, thanks for the explanation!

doc/crypto/appendix/encodings.rst
```diff
@@ -79,6 +81,7 @@ The CAT field in an algorithm identifier takes the values shown in :numref:`tabl
     MAC, ``0x03``, See :secref:`mac-encoding`
     Cipher, ``0x04``, See :secref:`cipher-encoding`
     AEAD, ``0x05``, See :secref:`aead-encoding`
+    Key wrapping, ``0x0B``, See :secref:`key-wrap-encoding`
```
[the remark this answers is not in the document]
athoelke: * Is a logical ordering algorithm categories in the operation documentation (chapter 10) the best approach (rather than appending new categories to the end)?
* Should the table here use that same ordering, or use the numerical category ordering?
* If these are different, should the encoding sub-sections (B.1.x) be ordered as in chapter 10, or as in the Algorithm category table here?

athoelke:
I have addressed Open issue 1 directly, and am happy with the encoding decisions

PR #230, opened 2024-12-05
Add definitions required by WPA3-SAE and WPA3-SAE-PT

oberon-sk:
The PSA Certified Crypto API v1.2 PAKE Extension requires some additional definitions to support WPA3-SAE and WPA3-SAE-PT as defined in IEEE standard specification [P802.11-REVme/D7.0, Part 11, Aug 2024](https://ieeexplore.ieee.org/document/10638490) to be embedded in, or included by, _psa/crypto.h_.

This PR provides a proposal.

### DIFF
--- a/headers/crypto/1.2/psa/crypto-wpa3-sae.h
+++ b/headers/crypto/1.2/psa/crypto-wpa3-sae.h
@@ -1,0 +1,313 @@
+/* This file contains reference definitions for implementations of 
+ * WPA3-SAE and WPA3-SAE-PT in addition to the PSA Certified Crypto
+ * API v1.2 PAKE Extension
+ *
+ * These definitions must be embedded in, or included by, psa/crypto.h
+ */
+ 
+#define PSA_KEY_TYPE_WPA3_SAE_PT_BASE          ((psa_key_type_t) 0x7800)
+#define PSA_KEY_TYPE_WPA3_SAE_GROUP_MASK       ((psa_key_type_t) 0x00ff)
+
+/** WPA3-SAE-PT key.
+ *
+ * The key is used to store the out of band calculated group element
+ * used in the Hash-To-Element variant of WPA3-SAE. It can be used as
+ * input to the WPA3-SAE PAKE instead of a password key.
+ *
+ * \param group A value of type ::psa_ec_family_t or ::psa_dh_family_t
+ *              that identifies the group to be used.
+ */
+#define PSA_KEY_TYPE_WPA3_SAE_PT(group) \
+    ((psa_key_type_t) (PSA_KEY_TYPE_WPA3_SAE_PT_BASE | (group)))
+
+ /** Whether a key type is a WPA3-SAE-PT. */
+#define PSA_KEY_TYPE_IS_WPA3_SAE_PT(type)                    \
+    (((type) & ~PSA_KEY_TYPE_WPA3_SAE_GROUP_MASK) ==         \
+     PSA_KEY_TYPE_WPA3_SAE_PT_BASE)
+ /** Extract the group from a WPA3-SAE key type. */
+#define PSA_KEY_TYPE_WPA3_SAE_PT_GET_FAMILY(type)            \
+    ((psa_ecc_family_t) (PSA_KEY_TYPE_IS_WPA3_SAE_PT(type) ? \
+      ((type) & PSA_KEY_TYPE_WPA3_SAE_GROUP_MASK) :          \
+      0))
+
+#define PSA_ALG_WPA3_SAE_PT_BASE          ((psa_algorithm_t) 0x08800400)
+/** The WPA3-SAE password to PT KDF.
+ * It takes the password p, a salt (uuid), and optionally a password id.
+ * 
+ * This key derivation algorithm uses the following inputs, which must be
+ * provided in the following order:
+ * - #PSA_KEY_DERIVATION_INPUT_SALT for the uuid.
+ * - #PSA_KEY_DERIVATION_INPUT_SECRET for the password.
+ * - optionally; #PSA_KEY_DERIVATION_INPUT_INFO for the password id.
+ * The output has to be read as a key of type PSA_KEY_TYPE_WPA3_SAE_PT.
+ *
+ * \param hash_alg      A hash algorithm (\c PSA_ALG_XXX value such that
+ *                      #PSA_ALG_IS_HASH(\p hash_alg) is true).
+ *
+ * \return              The corresponding counter-mode KDF algorithm.
+ * \return              Unspecified if \p hash_alg is not a supported
+ *                      hash algorithm.
+ */
+#define PSA_ALG_WPA3_SAE_PT(hash_alg)                            \
+    (PSA_ALG_WPA3_SAE_PT_BASE | ((hash_alg) & PSA_ALG_HASH_MASK))
+
+/** Whether the specified algorithm is a key derivation algorithm constructed
+ * using #PSA_ALG_WPA3_SAE_PT(\p hash_alg).
+ *
+ * \param alg An algorithm identifier (value of type #psa_algorithm_t).
+ *
+ * \return 1 if \p alg is a key derivation algorithm constructed using #PSA_ALG_WPA3_SAE_PT(),
+ *         0 otherwise. This macro may return either 0 or 1 if \c alg is not a supported
+ *         key derivation algorithm identifier.
+ */
+#define PSA_ALG_IS_WPA3_SAE_PT(alg)                         \
+    (((alg) & ~PSA_ALG_HASH_MASK) == PSA_ALG_WPA3_SAE_PT_BASE)
+
+/** The WPA3-SAE key exchange algorithm.
+ *
+ * This is WPA3-SAE as defined by "IEEE Std 802.11-REVme/D7.0 2024, 
+ * chapter 12.4", including the Hash-To-Element (H2E) variant.
+ * It is instantiated with the following parameters:
+ *
+ * - The group is defined over a finite field or an elliptic curve.
+ * - A cryptographic hash function.
+ *
+ * For WPA3-SAE selected by the AKM suites 8 and 9, use the
+ * PSA_ALG_WPA3_SAE_FIXED() algorithm, parameterized by the required hash
+ * algorithm.
+ * For WPA3-SAE selected by the AKM suites 24 and 25 (SAE using group-dependent
+ * hash), use the PSA_ALG_WPA3_SAE_GDH() algorithm, parameterized by the
+ * required hash algorithm.
+ * 
+ * To select these parameters and set up the cipher suite, call these functions:
+ *
+ * \code
+ * psa_pake_cipher_suite_t cipher_suite = PSA_PAKE_CIPHER_SUITE_INIT;
+ * psa_pake_cs_set_algorithm(cipher_suite, PSA_ALG_WPA3_SAE_FIXED(hash));
+ * psa_pake_cs_set_primitive(&cipher_suite,
+ *                           PSA_PAKE_PRIMITIVE(type, family, bits));
+ * \endcode
+ *
+ * After initializing a WPA3-SAE operation, call:
+ *
+ * \code
+ * psa_pake_setup(operation, password, cipher_suite);
+ * psa_pake_set_user(operation, ...);
+ * psa_pake_set_peer(operation, ...);
+ * \endcode
+ * 
+ * For basic SAE the password must be of type #PSA_KEY_TYPE_PASSWORD,
+ * for SAE-H2E the password must be of type #PSA_KEY_TYPE_WPA3_SAE_PT.
+ *
+ * \c psa_pake_set_role() must not be called because WPA3-SAE is a symmetric PAKE.
+ * 
+ * For a key exchange first call the following functions in any order:
+ * \code
+ * // send commit message
+ * psa_pake_output(operation, #PSA_PAKE_STEP_COMMIT, ...);
+ * // receive commit message
+ * psa_pake_input(operation, #PSA_PAKE_STEP_COMMIT, ...);
+ * \endcode
+ *
+ * If the Hash-To-Element variant is used and a list of rejected groups
+ * is available, it must be provided as a salt:
+ *
+ * \code
+ * // input salt
+ * psa_pake_input(operation, #PSA_PAKE_STEP_SALT, ...);
+ * \endcode
+ *
+ * Then call the following functions in any order:
+ * \code
+ * // set send-confirm counter
+ * psa_pake_input(operation, #PSA_PAKE_STEP_SEND_CONFIRM, ...);
+ * // send confirm message
+ * psa_pake_output(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // receive confirm message
+ * psa_pake_input(operation, #PSA_PAKE_STEP_CONFIRM, ...);
+ * // get key id (optional)
+ * psa_pake_output(operation, #PSA_PAKE_STEP_KEYID, ...);
+ * \endcode
+ * 
+ * Remarks:
+ * \c psa_pake_input(#PSA_PAKE_STEP_SEND_CONFIRM) must be called before
+ * \c psa_pake_output(#PSA_PAKE_STEP_CONFIRM) to set the send-confirm counter.
+ * The #PSA_PAKE_STEP_SEND_CONFIRM and #PSA_PAKE_STEP_CONFIRM steps may be used
+ * multiple times to handle repeated confirm messages with varying counts.
+ *
+ * Finally get the shared secret: 
+ *
+ * \code
+ * // get secret
+ * psa_pake_get_shared_key();
+ * \endcode
+ *
+ * The shared secret produced by WPA3-SAE is pseudorandom.
+ * It can be used directly as an encryption key or as input to a key derivation
+ * operation.
+ */
+#define PSA_ALG_WPA3_SAE_FIXED_BASE             ((psa_algorithm_t) 0x0a000800)
+#define PSA_ALG_WPA3_SAE_FIXED(hash_alg) (PSA_ALG_WPA3_SAE_FIXED_BASE | ((hash_alg) & PSA_ALG_HASH_MASK))
+#define PSA_ALG_WPA3_SAE_GDH_BASE               ((psa_algorithm_t) 0x0a000900)
+#define PSA_ALG_WPA3_SAE_GDH(hash_alg) (PSA_ALG_WPA3_SAE_GDH_BASE | ((hash_alg) & PSA_ALG_HASH_MASK))
+#define PSA_ALG_IS_WPA3_SAE(alg) (((alg) & ~0x000001ff) == PSA_ALG_WPA3_SAE_FIXED_BASE)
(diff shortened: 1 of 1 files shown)


athoelke:
The 2024 update to 802.11 is now finalized (late September) - I presume that contains all of the required specification now?

oberon-sk:
> The 2024 update to 802.11 is now finalized (late September) - I presume that contains all of the required specification now?

[IEEE Std 802.11™-2024 was approved on September 26, 2024. Publication expected soon. ](https://ieee802.org/11/)

athoelke:
Thank you for contributing these definitions. I would like to suggest some changes to the proposal:

1. A different name for the KDF used to calculate the password token from the password.
2. A different encoding of the password-token key types

### Password-token KDF name

In WPA3-SAE, the password token is computed using the hash-to-element procedure. My suggestion is to name the KDF after the procedure, rather than the result:

```
#define PSA_ALG_WPA3_SAE_H2E(hash_alg) ((psa_alogrithm_t) 0x08800400 | (0xff & hash_alg))
```

I presume the selection of the hash algorithm should follow the 802.11 specification, and match with the hash in the selected WPA3-SAE ciphersuite.

### Password token keys

There are two separate issues to address here:

1. The password token (PT) in WPA3-SAE is a group element, which is also the structure of an asymmetric public key in the same group. Which I guess is why the proposal encodes the password-token key as an asymmetric public key?

   However, for WPA3-SAE, the PT is not part of an asymmetric key, so a different key type encoding would be preferred.
But, unlike other key types in the Symmetric key category (`key_type & 0xf000 == 0x2000`), a PT key is not just unstructured bytes of key material, so that is also not preferred.

   I suggest that we allocate the currently unused encodings `key_type & 0xf000 == 0x3000` for 'structured keys' [that are not key-pairs or parts of key pairs].

   To be able to directly use the DH and ECC 'family' encodings, the final 12 bits would be split, similarly to Asymmetric keys, into a 5-bit structured-key-type value, a 6-bit family, and the 1-bit parity P bit.

2. The proposal allocates DH and ECC groups into the same key type. This just happens to work with the current specification because none of the allocated DH and ECC family values collide. However, the intention for these families is to be independently allocated, which includes the possibility of overlap in future.

   To preserve the intended independence, the password-token key types for DH-based WPA3-SAE and ECC-based WPA3-SAE need to have distinct encodings. To reuse the family encodings for ECC and DH keys, we need to match parity in the most-significant 9 bits with the ECC and DH key types. This means that the structured-key-type value must have an even number of 1-bits for both ECC and DH.

   I suggest we allocate a structured-key-type of 5 for ECC and 6 for DH:

   *  The ECC-based WPA3-SAE password token key encoding would be `0x3000 | (5 << 7) | ec_family`
   *  The DH-based WPA3-SAE password token key encoding would be `0x3000 | (6 << 7) | dh_family`
   
We also need to qualify the key type identifiers by group type. My suggestion would be:

```
#define PSA_KEY_TYPE_WPA3_SAE_ECC_PT(ec_family) ((psa_key_type_t)(0x3280 | (ec_family)))
#define PSA_KEY_TYPE_WPA3_SAE_DH_PT(dh_family)  ((psa_key_type_t)(0x3300 | (dh_family)))
```

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+ * For WPA3_SAE algorithms, the format for both input and output at this step
+ * is a 2 byte little-endian "send-confirm" counter followed by the output of
+ * the hash algorithm specified.
+ */
+#define PSA_PAKE_STEP_CONFIRM                   ((psa_pake_step_t)0x04)
+
+/** The WPA3-SAE commit step.
+ *
+ * The format for both input and output at this step is a 2 byte number
+ * specifying the group used followed by a scalar and an element of the
```
athoelke, line 177: Should we make the _commit-scalar_ and **_COMMIT_ELEMENT_** components of the commit frames separate outputs/inputs (as we do for the separate values in J-PAKE)?

Does concatenating them make it easier for the caller to assemble/disassemble the SAE Commit frames in 802.11?

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+ * The format of the value depends on the algorithm and cipher suite:
+ *
+ * For SPAKE2+ algorithms, the format for both input and output at this step is
+ * the same as the output of the MAC algorithm specified in the cipher suite.
+ *
+ * For PSA_ALG_SRP_6, the format for both input and output at this step is
+ * the same as the output of the hash algorithm specified.
+ *
+ * For WPA3_SAE algorithms, the format for both input and output at this step
+ * is a 2 byte little-endian "send-confirm" counter followed by the output of
```
athoelke, line 169: Although the prefix counter is redundant in the output, as the caller has just passed this as an input prior; it makes sense to keep this value symmetric for output and input, and it is required for input when verifying the peer's confirmation value using the peer's confirmation counter.

athoelke (changes requested):
I am starting to write this up for a draft PR against the specification, and a number of small details have emerged. Please see the the specific comments and questions.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+ * It takes the password p, a salt (uuid), and optionally a password id.
+ * 
+ * This key derivation algorithm uses the following inputs, which must be
+ * provided in the following order:
+ * - #PSA_KEY_DERIVATION_INPUT_SALT for the uuid.
+ * - #PSA_KEY_DERIVATION_INPUT_SECRET for the password.
+ * - optionally; #PSA_KEY_DERIVATION_INPUT_INFO for the password id.
+ * The output has to be read as a key of type PSA_KEY_TYPE_WPA3_SAE_PT.
+ *
+ * \param hash_alg      A hash algorithm (\c PSA_ALG_XXX value such that
```
athoelke, line 44: I am trying to decide if we need to parameterize the KDF, and the WPA3-SAE algorithms with a hash function at all. I think for implementations that follow the 80.11 spec precisely we could drop the explicit parameterization:

* The KDF implies the use of the hash-to-curve password token generation - and for this, the specification §12.4.2 is unambiguous about the mapping from cyclic group to hash function. So the hash function to use can be determined from the cyclic group specified in the output key type.
* When setting up the PAKE operation, the cipher suite specifies the cyclic group, and the type of key (PASSWORD or WPA_SAE_XX_PT) determines the choice of looping vs H2E generation of the PWE.

  - For looping, the hash function is fixed as SHA-256.
  - For H2E, the hash function is specified in Table 12-1. In this case, the PAKE set up must also verify that the key type matches the primitive in the cyclic group.

If we keep the parameterization, then we would either:

1. Require the implementation to verify that the hash matches the specification requirement. This is only valuable if we anticipate a future change in the standard to use other hash algorithms.
2. OR, we permit non-standard hash algorithms to be used with the selected cyclic group/PWE-generation method.

In either of those approaches, it becomes an application problem to select and specify the correct hash algorithm for the KDF and the PAKE operations. In (1), an incorrect value will result in a runtime error, in (2) it will result in an invalid, and possibly vulnerable authentication protocol.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+    (((alg) & ~PSA_ALG_HASH_MASK) == PSA_ALG_WPA3_SAE_PT_BASE)
+
+/** The WPA3-SAE key exchange algorithm.
+ *
+ * This is WPA3-SAE as defined by "IEEE Std 802.11-REVme/D7.0 2024, 
+ * chapter 12.4", including the Hash-To-Element (H2E) variant.
+ * It is instantiated with the following parameters:
+ *
+ * - The group is defined over a finite field or an elliptic curve.
+ * - A cryptographic hash function.
```
athoelke, line 73: See comment against L44 regarding the need to explicitly specify the hash parameter?

athoelke (changes requested):
Questioning the value in explicitly specifying the hash parameter.

An implementation that only supports the standard can determine the correct hash function from the cyclic group and PWE-generation method.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+      0))
+
+#define PSA_ALG_WPA3_SAE_PT_BASE          ((psa_algorithm_t) 0x08800400)
+/** The WPA3-SAE password to PT KDF.
+ * It takes the password p, a salt (uuid), and optionally a password id.
+ * 
+ * This key derivation algorithm uses the following inputs, which must be
+ * provided in the following order:
+ * - #PSA_KEY_DERIVATION_INPUT_SALT for the uuid.
+ * - #PSA_KEY_DERIVATION_INPUT_SECRET for the password.
```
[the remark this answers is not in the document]
oberon-sk: Agreed. We already use `PSA_KEY_DERIVATION_INPUT_PASSWORD` but the comment on `PSA_ALG_WPA3_SAE_PT` was wrong.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * - The group is defined over a finite field or an elliptic curve.
+ * - A cryptographic hash function.
+ *
+ * For WPA3-SAE selected by the AKM suites 8 and 9, use the
+ * PSA_ALG_WPA3_SAE_FIXED() algorithm, parameterized by the required hash
+ * algorithm.
+ * For WPA3-SAE selected by the AKM suites 24 and 25 (SAE using group-dependent
+ * hash), use the PSA_ALG_WPA3_SAE_GDH() algorithm, parameterized by the
+ * required hash algorithm.
```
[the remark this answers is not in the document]
oberon-sk: > Can/should we specify that if the selected hash does not match the one specified in §12.4.2 in the IEEE specification (SHA256 for generation of PWE by looping, following Table 12-1 for generation using hash-to-curve), the algorithm identifier will be rejected as invalid?

Agreed.

> This might be detected during password-token generation (the hash in the H2E KDF dos not match the cyclic group specified in the output key type); or during PAKE protocol flow if the hash WPA3-SAE algorithm identifier does not match with the cyclic group specified in the PAKE primitive.

We think it should be during password-token generation.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+ * psa_pake_cs_set_algorithm(cipher_suite, PSA_ALG_WPA3_SAE_FIXED(hash));
+ * psa_pake_cs_set_primitive(&cipher_suite,
+ *                           PSA_PAKE_PRIMITIVE(type, family, bits));
+ * \endcode
+ *
+ * After initializing a WPA3-SAE operation, call:
+ *
+ * \code
+ * psa_pake_setup(operation, password, cipher_suite);
+ * psa_pake_set_user(operation, ...);
```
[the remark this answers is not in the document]
oberon-sk: User and peer are 6 byte MAC addresses. But we have not found any reference to the format in the spec.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
```diff
@@ -0,0 +1,313 @@
...
+ * For a key exchange first call the following functions in any order:
+ * \code
+ * // send commit message
+ * psa_pake_output(operation, #PSA_PAKE_STEP_COMMIT, ...);
+ * // receive commit message
+ * psa_pake_input(operation, #PSA_PAKE_STEP_COMMIT, ...);
+ * \endcode
+ *
+ * If the Hash-To-Element variant is used and a list of rejected groups
+ * is available, it must be provided as a salt:
```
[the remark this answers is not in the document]
oberon-sk: Yes. Defined in 802.11 IEEE standard specification [P802.11-REVme/D7.0, Part 11, Aug 2024](https://ieeexplore.ieee.org/document/10638490) chapter 12.4.5.4, on page 3014.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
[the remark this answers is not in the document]
athoelke: Thank you for confirming

headers/crypto/1.2/psa/crypto-wpa3-sae.h
oberon-sk: We do not think it makes a big difference for the caller.
However, more STEP constants would be needed.

oberon-sk:
> Questioning the value in explicitly specifying the hash parameter.
> 
> An implementation that only supports the standard can determine the correct hash function from the cyclic group and PWE-generation method.

We used the hash parameter for consistency with the `PSA_ALG_WPA3_SAE_PT` KDF and other PAKE algorithms.
We agree it is not strictly necessary.

athoelke:
> > Questioning the value in explicitly specifying the hash parameter.
> > An implementation that only supports the standard can determine the correct hash function from the cyclic group and PWE-generation method.
> 
> We used the hash parameter for consistency with the `PSA_ALG_WPA3_SAE_PT` KDF and other PAKE algorithms. We agree it is not strictly necessary.

My inclination at this point is to remove the explicit parameterization - because it provides no benefit to the caller of the API, but does incur cost to the caller in determining what the correct value should be.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
oberon-sk: This is no problem for the PAKE operation.

For the KDF there is a problem regarding an efficient implementation:

The ECC or DH family is needed to fix the hash size. The family is provided as part of the key type contained in the attributes argument of the `psa_key_derivation_output_key()` function. This is the last call in a sequence of calls to KDF functions. This means the whole KDF calculations have to be delayed and executed in the `psa_pake_output_key` function because the Hash (and MAC) to be used is not known before. All inputs at the other steps have to be buffered in the operation object which is very undesirable for an efficient implementation.

This is why we strongly suggest to keep the hash argument in the KDF algorithm.

headers/crypto/1.2/psa/crypto-wpa3-sae.h
oberon-sk: see above

athoelke:
The effect of the incremental changes to WPA3-SAE (2016 -> 2020 -> 2024) is that there are notionally 3 distinct variants:

1. PWE generation by looping, using SHA-256 as H() for all cipher suites. len(PMK) = len(KCK) = 256
2. PWE generation using hash-to-element, determining H() from the size of the group prime. len(PMK) = 256. len(KCK) = len(H())
3. PWE generation using hash-to-element, determining H() from the size of the group prime. len(PMK) = len(KCK) = len(H())

At the moment, the proposal defines one algorithm for (1) and (2), differentiating by the key type of the key that is passed in to set up the PAKE operation. And a second algorithm for (3), to separate it from (2).

Would the API be clearer if we gave all three their own algorithm identifier instead?

athoelke:
The 802.11-2024 specification suggests that for devices might support both looping and H2E methods for generating the PWE - and the method selected can vary between authentication instances.

That means the password must be usable with both the `PSA_ALG_WPA3_SAE_FIXED` algorithm in a PAKE operation (for looping method); and with the `PSA_ALG_WPA3_SAE_H2E` algorithm in a key derivation operation. A key of type `PSA_KEY_TYPE_PASSWORD`, cannot permit both of these algorithms. Do we need a wildcard algorithm for these keys?

A `PSA_KEY_TYPE_WPA3_SAE_XX_PT` key might be used with both `PSA_ALG_WPA3_SAE_FIXED` and `PSA_ALG_WPA3_SAE_GDH`, as these only differ in the size of the output PMK (and thus the KCK and PMK values). Do we need a wildcard algorithm for the password token keys?

headers/crypto/1.2/psa/crypto-wpa3-sae.h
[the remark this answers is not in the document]
athoelke: > We think it should be during password-token generation

Yes. I think I meant to say that the verification should occur in *both* places, not a choice of one or the other.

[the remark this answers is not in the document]
athoelke: But this is only true if we keep the parameterization (see later comments)

headers/crypto/1.2/psa/crypto-wpa3-sae.h
athoelke: Ok - a valid counter-argument - the HKDF-Extract phase could consume the other inputs as they are provided - as long as we know the Hash function from the start.

athoelke:
> My inclination at this point is to remove the explicit parameterization - because it provides no benefit to the caller of the API, but does incur cost to the caller in determining what the correct value should be.

Spoke too soon - before you explained the benefit of knowing the Hash to start computing HKDF prior to knowing the cyclic group from the output key type.

oberon-sk:
> > My inclination at this point is to remove the explicit parameterization - because it provides no benefit to the caller of the API, but does incur cost to the caller in determining what the correct value should be.
> 
> Spoke too soon - before you explained the benefit of knowing the Hash to start computing HKDF prior to knowing the cyclic group from the output key type.

The hash is strictly required only in the KDF algorithm. However, we would prefer if the parameterization would be used in both, the KDF algorithm and the PAKE algorithm.

oberon-sk:
> The 802.11-2024 specification suggests that for devices might support both looping and H2E methods for generating the PWE - and the method selected can vary between authentication instances.
> 
> That means the password must be usable with both the `PSA_ALG_WPA3_SAE_FIXED` algorithm in a PAKE operation (for looping method); and with the `PSA_ALG_WPA3_SAE_H2E` algorithm in a key derivation operation. A key of type `PSA_KEY_TYPE_PASSWORD`, cannot permit both of these algorithms. Do we need a wildcard algorithm for these keys?
> 
> A `PSA_KEY_TYPE_WPA3_SAE_XX_PT` key might be used with both `PSA_ALG_WPA3_SAE_FIXED` and `PSA_ALG_WPA3_SAE_GDH`, as these only differ in the size of the output PMK (and thus the KCK and PMK values). Do we need a wildcard algorithm for the password token keys?

Agreed. Both wildcards are needed to allow to use stored passwords and PT keys in all cases.

athoelke:
I've now turned this into a complete documention PR (#293), which tries to include all of the contribution and discussion from here.

The result is similar to this proposal, but has the following differences:

1. The password token keys are a new category of key: 'structured'. They are not key pairs, so this approach makes better use of key type encoding space.
2. ECC and DH groups are encoded in a single family encoding - so to avoid future issues, there are separate password token keys for ECC and DH primitives.
3. The KDF for hash-to-element has a different name `PSA_ALG_WPA3_SAE_H2E()`.
4. The commit values are output and input from the operation as separate steps, `PSA_PAKE_STEP_KEY_SHARE` is reused for the element values, `PSA_PAKE_STEP_COMMIT_SCALAR` is added for the scalar.

There are some open issues, including the handling of the commit values, the naming of the key types, and whether wildcard permitted-algorithm key policies are needed for the WPA3-SAE use cases.

PR #233, opened 2024-12-19
Add support for stateful hash-based signatures to the PQC Extension

### LINKED ISSUE #232: Support for stateful-hash-based signature verification (HSS/LMS, XMSS, and XMSS^MT)
Prior to the standardization of ML-DSA and SLH-DSA, stateful-hash-based signature algorithms were developed, and then described in [NIST Special Publication 800-208](https://doi.org/10.6028/NIST.SP.800-208). The algorithms are:

* LMS and its multi-level variant HSS/LMS. These are described in [RFC 8554](https://www.rfc-editor.org/info/rfc8554)
* XMSS and its multi-level variant XMSS<sup>MT</sup>. These are described in [RFC 8391](https://www.rfc-editor.org/info/rfc8391)

These schemes are not vulnerable to quantum computing, and are based on the well-established security of cryptographic hash algorithms. They also have relatively small public keys and signatures (for PQC signature schemes), and relatively simple implementation requirements. However, they are based on one-time-signatures (OTS), and the security critically depends on never reusing any of the set of OTS private keys. This requirement is met by having a stateful private key, that records which OTS private keys have been used.

The implication of a stateful private key, is that deploying systems which use them have substantial system challenges related to throughput, resilience and redundancy. The security and footprint attributes of these schemes (compared to other PQC signatures) make them attractive for some specific use cases at this point in time. For example, authentication of initial stages of constrained system firmware, where the authenticating component and authentication public key are immutable.

The use case of firmware verification can be supported without needing an implementation to deal with the challenge of handling a stateful private key: the Crypto API could define suitable public-key types and signature algorithms, that can only be used with signature verification functions.

athoelke:
Add support for the stateful hash-based signature schemes LMS, HSS, XMSS, and XMSS<sup>MT</sup> to the PQC Extension.

This support is limited to public keys and signature verification - key generation and secure management of the stateful private key is not within the current scope of this proposal.

As the key types and signatures use XDR structures which encode the algorithm parameterisation, and public keys must be imported - there is no need to parameterise the key types or the algorithm identifiers in the Crypto API. A single key type and algorithm identifier is allocated for each of the three schemes.

An updated rendering of the extension document is attached: 
[AES0119-PSA_Certified_Crypto_API-1.3_PQC_Extension-bet.0-draft4.pdf](https://github.com/user-attachments/files/18690560/AES0119-PSA_Certified_Crypto_API-1.3_PQC_Extension-bet.0-draft4.pdf)

### Open Issues

* `key_bits` values to select between the ML-DSA-44/65/87 parameter sets. Should we use the literal values in the name, or a logical lambda strength value as suggested, or something else?

Fixes #232


athoelke:
Thank you for the feedback.

I had not thought about scenarios where an application might be using LMS independently of X.509, or related protocols. It would indeed be awkward to use the proposed API in that situation.

I will add identifiers for LMS public keys and the LMS signature algorithm. I think the suggestion to rename the HSS/LMS identifiers to be just HSS makes sense as well.

athoelke:
I've updated the PR, including a separate key type and algorithm identifier for LMS, when not used in HSS. This supports use cases which are not using X.509 infrastructure or key formats, and dynamic conversion to HSS is impractical.

I have also implemented a suggestion for the use of `key_bits` in these key types. For all of these keys, the formatted key size is a simple calculation based on the size of the hash used. My initial suggestion is that `key_bits` should equal the hash output size (e.g. 192 or 256 for the NIST-approved parameter sets in SP800-208). Alternatively, we could use the actual encoded key size in bits. In practice, this value can be left as zero when importing a key, as `key_bits` will be deduced from the imported key data.

Unless there is a solid request for it, I will keep separate identifiers for the four algorithms.

athoelke:
Updated draft(3) PDF attached to PR description.

oberon-sk:
> I have proposed having the key_bits in the key attributes match the output size of the hash function that parameterises the signature scheme. This seems more intuitive to use than the actual key sizes.

Agreed.

oberon-sk:
Regarding [AES0119-PSA_Certified_Crypto_API-1.3_PQC_Extension-bet.0-draft3.pdf](https://github.com/user-attachments/files/18456705/AES0119-PSA_Certified_Crypto_API-1.3_PQC_Extension-bet.0-draft3.pdf),
2.3.1 Module Lattice-based signature keys:

> [FIPS204]. The values are based on the dimensions of the matrix 𝐴, and do not directly define the key size
> in bytes:
> ∙ ML-DSA-44 : key_bits = 44
> ∙ ML-DSA-65 : key_bits = 65
> ∙ ML-DSA-87 : key_bits = 87

Instead of 44, 65, 87, we would prefer something more expressive. The lambda value would be a candidate as it roughly indicates the security levels: 128, 192, 256.

athoelke:
The stateful hash-based signatures (HSS,LMS, XMSS and XMSS<sup>MT</sup>) can be parameterised with four different hash algorithms (SHA-256, SHA-256/192, SHAKE256/256, and SHAKE256/192).

The signature schemes do not permit pre-hashing of the message, and do not require the application to use a hash algorithm identifier as the hash-function is fixed at key generation time and we are not supporting generation of keys.

However, there is value in defining identifiers for the three that are missing from the API, for use within API implementations that integrate multiple cryptographic components.

athoelke:
Added missing hash algorithms, and removed Todos from the document.

Updated draft PDF attached to the PR description.

athoelke:
> Regarding [AES0119-PSA_Certified_Crypto_API-1.3_PQC_Extension-bet.0-draft3.pdf](https://github.com/user-attachments/files/18456705/AES0119-PSA_Certified_Crypto_API-1.3_PQC_Extension-bet.0-draft3.pdf), 2.3.1 Module Lattice-based signature keys:
> 
> > [FIPS204]. The values are based on the dimensions of the matrix 𝐴, and do not directly define the key size
> > in bytes:
> > ∙ ML-DSA-44 : key_bits = 44
> > ∙ ML-DSA-65 : key_bits = 65
> > ∙ ML-DSA-87 : key_bits = 87
> 
> Instead of 44, 65, 87, we would prefer something more expressive. The lambda value would be a candidate as it roughly indicates the security levels: 128, 192, 256.

I've given more thought to this approach, and I think I am inclined to agree. Using these notional 'security strength' values is also more consistent with both the API in general, and also with the specific key-bits sizes used for other PQC signature algorithms SLH-DSA, LMS, and XMSS.

PR #243, opened 2025-03-03
Change the bit size used to select a specific ML-DSA key parameter set

### LINKED ISSUE #241: Change the `key_bits` values for ML-DSA parameter sets to be the security strength
In the current draft PQC Extension, the parameter-set value (44, 65, or 87) is used as the `key_bits` value in the key attributes to select a specific ML-DSA parameter set.

As [commented here](https://github.com/ARM-software/psa-api/pull/233#issuecomment-2684663059) in the related PR for stateful hash-based signatures; it would be more consistent with the other algorithms for ML-DSA to use a 'security strength' value for `key_bits`. I.e. use the values 128, 192, and 256 instead.

athoelke:
Base the key size on the security (collision) strength of the ML-DSA parameter set, rather than on the name of the parameter set.

Fixes #241


gilles-peskine-arm (approved):
LGTM

PR #253, opened 2025-04-30
Add support for XOF algorithms

### LINKED ISSUE #248: API for XOF algorithms
It probably time to provide support for XOF algorithms in the Crypto API. We now have SHAKE128 and SHAKE256, and the being-finalised Ascon-XOF128 and Ascon-CXOF128 XOF algorithms.

Although the dominant use of these algorithms is in the implementation of more complex algorithms, such as ML-DSA and ML-KEM, or as fixed-size-output hashes (e.g. SHAKE128/256), a software implementation of the latter algorithms will want to use an accelerated implementation of the XOF. For this, we need to define some API for accessing the XOF algorithm directly.

There are three ways to approach this:

1. Extend the multi-part Hash operation, to permit repeated extraction of the output, and accept an optional context input.
2. Standardise a use of the key derivation operation for XOF algorithms, which already supports incremental input (at least for the SALT input in PBKDF2-HMAC) and output.
3. Define a new XOF operation API.

(1) is not a really a good fit:

* Hashes are one of the most used, and simplest, cryptographic primitives. The API should not be complicated by support for adjacent algorithm types.
* We don't want implementations to have to support incremental output of hash algorithms. The functions that can be used on a hash operation will depend on the algorithm selected - losing API agility.
* Fixed-size output from an XOF is a Hash (e.g. SHAKE128/256), but where this is important, such as a hashing option for SLH_DSA, we need to define a separate Hash algorithm that reports the expected value for `PSA_HASH_LENGTH()`.

(2) was the tentative conclusion of earlier discussions - this was referenced in [the Ascon issue](https://github.com/ARM-software/psa-api/issues/60#issuecomment-2479023116). Since then, some additional considerations have come to light that might recommend an alternative approach for XOF support. In particular:

* Although some XOF algorithms can be used effectively as a KDF, without needing an HKDF-style construction, it is not intuitive that the key derivation API is what should be used for XOF.
* Filtering out and validating XOF usage in an implementation of the complex key derivation API is not good for the use case of using an accelerated XOF within a more complex software algorithm.

For both application code readability, and implementation optimisation, it is worth considering (3) creating a new dedicated API for XOF:

* A multi-part operation type, that accepts multiple inputs, an optional context, and can then generate multiple outputs, before being explicitly terminated.
* A distinct category of algorithms, initially for the SHAKE and Ascon variants.

As with the other Crypto APIs, the inputs and outputs would be octet/byte-based. I.e. the input can only be provided, and output can only be extracted, in whole numbers of bytes.

I do not think there is a strong argument to also add a single-part function for computing an XOF output in one go, from a single buffer of input?

athoelke:
* Add `psa_xof_operation_t` and its functions.
* Add SHAKE128 and SHAKE256 XOF algorithms.

Fixes #248


athoelke:
Would value any feedback on the naming of the input/output functions for the XOF operation API. As noted in the TODO in this PR:

> Should it be `psa_xof_input()` and `psa_xof_output()` (current), or `psa_xof_input_bytes()` and `psa_xof_output_bytes()`, or something else?
>
> For comparison:
>
> *   Hash operations use ``update()`` and ``finish()``, but only have a single output call. ``update()`` and ``output()`` doesn't seem quite right.
> *   KDF operations have ``input_bytes()`` and ``output_bytes()``, but also have ``_key()`` variants so distinguishing bytes vs keys is necessary for KDF.
> *   PAKE operations has ``input()`` and ``output()`` for the data transfer protocol steps.

MarcusJGStreets (approved):
I think I like psa_xof_input() and psa_xof_output().

As a XOF does not accept a key, the _bytes() seems tautologous to me, and therefore I would go for the shorter name. 

So I am happy as is - though obviously the todo section needs to be removed. 

I like the change to numbering in tens - reminds me of BBC BASIC. :-)

athoelke:
Rebased after fixing copyright dates in one of the set-up commits.

athoelke:
I am tempted to revisit the naming convention question.

All existing multi-part operations that take incremental inputs use the verb 'update' to add inputs: hash, MAC, cipher, AEAD. KDF uses the verb 'input' as there are many different inputs, and they are not streamed into the KDF.

For streamed output, KDF does use 'output', but in this case, the behavior of the output functions is the same as we need for XOF.

So, my revised opinion is to use the pair: `psa_xof_update()` and `psa_xof_output()`. `psa_xof_update()` matches expectation for a developer familiar with the existing multi-part operations.

Although this is not 'a matching pair', that statement is true of all of the other streaming input multi-part operations, which use update/finish or update/compare.

athoelke:
Changed the input function to `psa_xof_update()` to align with other multi-part operations in the API.

PR #260, opened 2025-05-13
Fix incorrect typename in custom key creation functions

### LINKED ISSUE #259: mbedTLS typename incorrectly used in Crypto API 1.3.0
When defining the custom key generation and key derivation APIs, the content sourced from the MbedTLS experimental implementation of the API was not fully cleaned up for use as specification source.

Two instances of the `mbedtls_svc_key_id_t` type were used in the API definition instead of `psa_key_id_t`.

[A search for 'medtls_svc_key_id_t'](https://arm-software.github.io/psa-api/crypto/1.3/search.html?q=mbedtls_svc_key_id_t) shows the two instances.

athoelke:
Change `mbedtls_svc_key_id_t` to `psa_key_id_t` in two function parameters introduced in version 1.3.

Fixes #259


gilles-peskine-arm (approved):
Thanks for noticing and fixing this

PR #261, opened 2025-05-13
Define `psa_attach_key()` to register existing key material with a key id and policy

### LINKED ISSUE #242: API function to register an existing key
As the developer of an application (or application stack) using the PSA crypto API on a device with a secure element,
I want to access keys that are provisioned onto the device by a process outside of my control,
so that my application can use pre-provisioned keys while sticking to the PSA API.

The idea is that many devices that have a secure element have a factory process for injecting credentials. That process exists and can't be retooled easily, so we can't insist that this process becomes aware of the PSA API. The device does have a software stack that implements the PSA crypto API. Through that API, we can create new keys and use them. But there's no standard way to use the credentials that are provisioned during the device manufacturing.

More generally, there's interest in _using_ a key through PSA even when it's created through proprietary means.

Mbed TLS implemented a function to do that alongside its original secure element interface: [`mbedtls_psa_register_se_key()`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.6.2/include/psa/crypto_extra.h#L148). That API was tied to the interface, and had some limitations (notably, [it can't be used with volatile keys](https://github.com/Mbed-TLS/mbedtls/issues/9254)), so in our next major release (TF-PSA-Crypyo 1.0), [we are evolving the API](https://github.com/Mbed-TLS/mbedtls/issues/9255). We have heard interest in making this feature official in the PSA API.

The starting proposal is as follows:

```
psa_register_opaque_key(
    const psa_key_attributes_t *attributes,
    const uint8_t *label, size_t label_length,
    psa_key_id_t *key_id);
```

This is a key creation function, taking `attributes` as input and `key_id` as output like the others. The way to obtain the key material is by associating the PSA key identifier with some existing “slot”. In PSA terms, the meaning of the `label` parameter is implementation-specified, and likely to be different for each supported key location.

With Mbed TLS opaque drivers, there are two plausible meanings for `label`:

* It could be whatever the driver expects as key data — what you'd get as output from asking an opaque driver to import/generate/derive a key.
* Or it could be an input passed to a driver callback.

In practice, the primary intent is to pass a slot number, or slot label, or key name, in the secure element.

Unresolved question: does `psa_destroy_key` only destroy the PSA key stub, or does it also destroy the underlying key?

Unresolved question: what can this function mean for transparent keys (keys in the location `PSA_KEY_LOCATION_LOCAL_STORAGE`)? Possibly this might duplicate the functionality that Mbed TLS calls [“built-in keys”](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.6.2/include/psa/crypto_extra.h#L478), which are a way to access keys that were not created through PSA, such as a HUK or HUK-derived key. These are independent features in Mbed TLS partly for historical reasons, but also because they are intended for different scenarios, and it's unclear whether a unified API would handle all the desired scenarios. Typically, the set of available built-in keys is decided by the device integrator and will not change over time, whereas registrable keys can be created dynamically if the proprietary provisioning scheme extends after the factory phase of the device lifecycle. Also they interact very differently with driver dispatch.

Timeline: we would ideally like to release the official `psa_register_opaque_key` in TF-PSA-Crypto 1.0, which means the official API should be stable (not necessarily released) by late June 2025.

athoelke:
This follows the discussion in #242. I think I've captured the end state of that discussion, please review and let me know what I have missed.

I still need to do some clarification rework relating to the meaning of 'volatile key' - but that should be first addressed as a fix to 1.3, and then ported into this work that targets 1.4.0.

Fixes #242

### DIFF
--- a/doc/crypto/api/keys/management.rst
+++ b/doc/crypto/api/keys/management.rst
@@ -460,7 +460,7 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
     .. retval:: PSA_SUCCESS
         Success.
     .. retval:: PSA_ERROR_DOES_NOT_EXIST
-        `label`` does not refer to key material within the location specified in ``attributes``.
+        ``label`` does not refer to key material within the location specified in ``attributes``.
     .. retval:: PSA_ERROR_NOT_SUPPORTED
         The key attributes, as a whole, are not supported, either by the implementation in general or in the specified storage location.
     .. retval:: PSA_ERROR_INVALID_ARGUMENT
--- a/doc/crypto/api/keys/management.rst
+++ b/doc/crypto/api/keys/management.rst
@@ -21,6 +21,7 @@ New keys can be created in the following ways:
 *   `psa_encapsulate()` and `psa_decapsulate()` create a shared secret key using a key-encapsulation mechanism.
 *   `psa_pake_get_shared_key()` creates a key from the shared secret result of a password-authenticated key exchange. See :secref:`pake`.
 *   `psa_copy_key()` duplicates an existing key with a different lifetime or with a more restrictive usage policy.
+*   `psa_attach_key()` registers existing key material from secure storage for use as a volatile key.
 
 When creating a key, the attributes for the new key are specified in a `psa_key_attributes_t` object. Each key creation function defines how it uses the attributes.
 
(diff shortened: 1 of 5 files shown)


athoelke:
I realise that the wording of the API definition gives the impression that this is designed for already-stored key material, but does not suggest that the API is valid and useful for key-ladder scenarios where a key can be repeatedly derived if and when required from a bult-in secret.

athoelke:
Rebased after fixing copyright dates in one of the set-up commits.

athoelke:
Added a new commit to include the use case for the API where the key material is derived from an internal secret (e.g. HUK), and not literally stored in the device. This for example could be used for device-bound, application-specific derivation keys, and could optional use some or all of the `label` input in the key derivation process.

athoelke:
> But there is also the case where the type is valid for the implementation but not supported by this location.

I think this scenario is covered by the text for `PSA_ERROR_NOT_SUPPORTED`: "The key attributes, as a whole, are not supported, either by the implementation in general or in the specified storage location."

"The key type is invalid" means that the key type is not recognised as a defined key type value. I don't think we needed to specify additional constraints on which types are allowed for this API (unlike, for example key generation not being permitted for asymmetric public keys).

doc/crypto/api/keys/management.rst
```diff
@@ -412,6 +413,114 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
+
+    This function allows applications to register existing key material that has been provisioned outside the |API|, such as during manufacturing or by a secure element.
+    It can also be used to create a key identifier for keys that are deterministically derived from a secret within the implementation.
+    After registering the key, the application has a volatile key identifier that can be used in cryptographic operations permitted by its usage flags and algorithm policy.
+
+    The key material is identified by its location, specified in the provided attributes ``lifetime`` value, and the ``label`` parameter.
+    The format of the label is specific to the implementation and storage location.
+    Typically, the label is used as a location-specific identifier for the key material, or can provide input for deriving the key material from an internal secret.
+
+    This function can only be used to create a volatile key.
```
gilles-peskine-arm, line 497: Somewhere (maybe not here) we should state that the behavior with persistent lifetimes is reserved and may be specified in a future version of the specification, to avoid divergent implementation-specific extensions.

gilles-peskine-arm (changes requested):
Looks mostly good to me, apart from some minor points around attributes.

doc/crypto/api/keys/management.rst
```diff
@@ -412,6 +413,114 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
+
+        The attributes that are required to be set depend on the identified key material.
+        Where an attribute is provided by both the application and the key material, they must be compatible.
+
+        The following attributes must be set:
+
+        *   The key lifetime must specify a volatile key, and the storage location of the existing key material.
+            See :secref:`key-lifetimes`.
+        *   The key identifier must be the default `PSA_KEY_ID_NULL`.
+        *   The key type.
```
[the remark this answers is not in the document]
athoelke: > Rather than distinguish between attributes that “must be set” and attributes that are “optional”, it would be more generic to specify that all attributes must be provided, either by the implementation or by the application, and that the set of attributes provided by the implementation is implementation-defined and possibly key-specific.

I was trying to keep this entirely consistent with the other key-creation functions. However, the comments you raise reflect my own struggle with that approach. I will rework it as suggested (which is a pattern we are likely to reuse for non-default import formats).

doc/crypto/api/keys/management.rst
athoelke: I think perhaps an Implementation Note here might work

doc/crypto/api/keys/management.rst
```diff
@@ -412,6 +413,114 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
+    .. summary::
+        Register existing key material within the implementation with a volatile key identifier and key policy.
+
+        .. versionadded:: 1.4
+
+    .. param:: const psa_key_attributes_t * attributes
+        The attributes for the key to be registered.
+
+        The attributes that are required to be set depend on the identified key material.
+        Where an attribute is provided by both the application and the key material, they must be compatible.
```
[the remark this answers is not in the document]
athoelke: I've reworded the API definiton to describe 'implementation-provided key [material]` that works better for the range of use cases we want this API to cover.

doc/crypto/api/keys/management.rst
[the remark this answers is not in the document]
athoelke: Reworked this section

doc/crypto/api/keys/management.rst
```diff
@@ -412,6 +413,114 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
+
+    It is :scterm:`implementation defined` whether the same key material can be attached to multiple key identifiers concurrently.
+
+    .. note::
+
+        This function is intended for scenarios where key material is provisioned outside the |API|, and the application needs to use such keys within the API framework.
+
+        Implementations may impose restrictions on which keys can be registered, depending on their storage architecture and security policies.
+
+        The function does not allow registering keys that are not already present in the implementation's storage.
```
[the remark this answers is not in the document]
athoelke: Reworked the notes

doc/crypto/api/keys/management.rst
```diff
@@ -497,30 +496,38 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
+    *   Provisioned key material has a fixed size.
+        The implementation might permit the application to define the key type and policy, as long as these are compatible with the key material.
+    *   An implementation-specific derived key can require the application to provide a key type and size, using these in the derivation process.
+    *   An implementation-provided key can be fully defined by the implementation, with a fixed type, size, and policy.
+        The call to `psa_attach_key()` needs to specify the location and label of the key, and a macthing policy, in order to obtain a key id.
 
-    Calling `psa_destroy_key()` with a key identifier returned by `psa_attach_key()` will remove the key identifer and policy from the key store, but the original key material remains within the implementation.
+    Calling `psa_destroy_key()` with a key identifier returned by `psa_attach_key()` will remove the key identifer and policy from the key store, but any implementation-provided key material remains within the implementation.
```
gilles-peskine-arm, line 508: Typo: identifer → identifier

doc/crypto/api/keys/management.rst
```diff
@@ -497,30 +496,38 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
 
     .. note::
 
-        This function is intended for scenarios where key material is provisioned outside the |API|, and the application needs to use such keys within the API framework.
+        This function is intended for scenarios where key material is provided outside the |API|, and the application needs to use such keys within the |API| framework.
 
-        Implementations may impose restrictions on which keys can be registered, depending on their storage architecture and security policies.
-
-        The function does not allow registering keys that are not already present in the implementation's storage.
+        The function does not allow registering keys that are not already provided by the implementation.
```
gilles-peskine-arm, line 517: I find this sentence hard to understand. The _key material_ is provided by the implementation but the _key object_ isn't. Also, there's a double negation, which is rarely easy to read. Suggestion:

> The function only allows registering key material that is provided by the implementation. To import new key material, use `psa_import_key()`.

doc/crypto/api/keys/management.rst
```diff
@@ -497,30 +496,38 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
...
+        Although the implementation verifies that the application-supplied attributes are compatible with the implementation-provided key; it is the application's responsibility to ensure correctness for attributes that are provided by the implementation.
 
         To create a persistent key from pre-existing key material, the implementation might permit a key returned by `psa_attach_key()` to be copied to a persistent key using `psa_copy_key()`.
 
+    .. admonition:: Implementation note
+
+        Implementations may impose restrictions on which keys can be registered, depending on their storage architecture and security policies.
+
+        The behavior of a call `psa_attach_key()` with a persistent key-lifetime might be specified in a future version of the |API|.
+        Such a call must result in a :code:`PSA_ERROR_INVALID_ARGUMENT` error in an implementation of version |docversion| of the |API|.
```
gilles-peskine-arm, line 529: Stating that the call *must* return this error means that if we define the behavior in 1.5, it'll be impossible for an implementation to claim both 1.5 and 1.4 compatibility. I don't think that's desirable. Also, an implementation might want to return a different error code, for example an implementation without persistent storage might return `PSA_ERROR_NOT_SUPPORTED` for any access to persistent storage. So I think this should be a SHOULD, or reworded to say that implementations should return an error rather than defining implementation-specific behavior.

gilles-peskine-arm (approved):
LGTM

doc/crypto/api/keys/management.rst
athoelke: Yes, removing the double negative is much clearer.

doc/crypto/api/keys/management.rst
athoelke: I think that was your original wording in a comment - but we should relax this.

athoelke: Here's my proposed alternative wording:

> The behavior of a call `psa_attach_key()` with a persistent key-lifetime might be specified in a future version of the Crypto API. At present, it is recommended that such a call returns an error status, and does not provide implementation-specific behavior.

PR #264, opened 2025-05-15
Contexts in signature algorithms

MarcusJGStreets:
In initial draft for four next functions to cover signing and verification with a non-empty context string.

### DIFF
--- a/doc/crypto/api/ops/signature.rst
+++ b/doc/crypto/api/ops/signature.rst
@@ -84,6 +84,12 @@ The |API| provides several functions for calculating and verifying signatures:
 
     If called with a zero-length context, these functions produce the same signature as the associated function without a context parameter.
 
+    .. note::
+
+        If a signature scheme treats the absence of a context parameter differently to a zero-length context, the |API| defines distinct algorithm identifiers for the two variants.
+        For example, when using a 255-bit key with EdDSA, `PSA_ALG_PURE_EDDSA` implements Ed25519 (without a context) and `PSA_ALG_EDDSA_CTX` implements Ed25519ctx (with a context, which can be zero-length).
+        See :secref:`eddsa-sign-algorithms`.
+
     It is an error to provide a non-zero-length context with an algorithm that does not accept contexts.
 
     Code written to be cryptographically agile can use the new functions, provided it guards against providing a non-zero-length context with an algorithm that does not support them.


doc/crypto/api/ops/signature.rst
```diff
@@ -629,6 +626,72 @@ Asymmetric signature functions
...
+        The key does not have the `PSA_KEY_USAGE_SIGN_MESSAGE` flag, or it does not permit the requested algorithm.
+    .. retval:: PSA_ERROR_BUFFER_TOO_SMALL
+        The size of the ``signature`` buffer is too small.
+        `PSA_SIGN_OUTPUT_SIZE()` or `PSA_SIGNATURE_MAX_SIZE` can be used to determine a sufficient buffer size.
+    .. retval:: PSA_ERROR_NOT_SUPPORTED
+        The following conditions can result in this error:
+
+        *   ``alg`` is not supported, or is not an asymmetric signature algorithm that permits signing a message.
+        *   ``key`` is not supported for use with ``alg``.
+        *   ``input_length`` is too large for the implementation.
```
athoelke, line 674: Add items for:
* Contexts are not supported by the implementationfor the algorithm.
* \`\`context_length\`\` is too large for the implementation.

doc/crypto/api/ops/signature.rst
```diff
@@ -629,6 +626,72 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The following conditions can result in this error:
+
+        *   ``alg`` is not an asymmetric signature algorithm that permits signing a message with a context string.
+        *   ``key`` is not an asymmetric key pair, that is compatible with ``alg``.
+        *   ``input_length`` is too large for the algorithm and key type.
```
athoelke, line 680: Add items for:
* Contexts are not valid with this algorithm.
* \`\`context_length\`\` is too large for the algorithm.

doc/crypto/api/ops/signature.rst
```diff
@@ -629,6 +626,72 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_CORRUPTION_DETECTED
+    .. retval:: PSA_ERROR_STORAGE_FAILURE
+    .. retval:: PSA_ERROR_DATA_CORRUPT
+    .. retval:: PSA_ERROR_DATA_INVALID
+    .. retval:: PSA_ERROR_INSUFFICIENT_ENTROPY
+    .. retval:: PSA_ERROR_BAD_STATE
+        The library requires initializing by a call to `psa_crypto_init()`.
+
+    ..  note::
+        To perform a multi-part hash-and-sign signature algorithm, first use a :ref:`multi-part hash operation <hash-mp>` and then pass the resulting hash to `psa_sign_hash()`. :code:`PSA_ALG_GET_HASH(alg)` can be used to determine the hash algorithm to use.
```
athoelke, line 692: Here, this would need to refer to `psa_sign_hash_with_context()`.

doc/crypto/api/ops/signature.rst
```diff
@@ -681,6 +744,65 @@ Asymmetric signature functions
     ..  note::
         To perform a multi-part hash-and-sign signature verification algorithm, first use a :ref:`multi-part hash operation <hash-mp>` to hash the message and then pass the resulting hash to `psa_verify_hash()`. :code:`PSA_ALG_GET_HASH(alg)` can be used to determine the hash algorithm to use.
 
+
+.. function:: psa_verify_message_with_context
+
+    .. summary::
+        Verify the signature of a message with a public key and a non-default context. For hash-and-sign algorithms, this includes the hashing step.
```
athoelke, line 751: Do we want to refer to a 'non-default' context? Should we say 'non-empty' context instead - or can we anticipate that a default would ever be different from an empty string?

doc/crypto/api/ops/signature.rst
```diff
@@ -681,6 +744,65 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_NOT_PERMITTED
+        The key does not have the `PSA_KEY_USAGE_VERIFY_MESSAGE` flag, or it does not permit the requested algorithm.
+    .. retval:: PSA_ERROR_INVALID_SIGNATURE
+        ``signature`` is not the result of signing the ``input`` message with algorithm ``alg`` using the private key corresponding to ``key``.
+    .. retval:: PSA_ERROR_NOT_SUPPORTED
+        The following conditions can result in this error:
+
+        *   ``alg`` is not supported, or is not an asymmetric signature algorithm that permits verifying a message.
+        *   ``key`` is not supported for use with ``alg``.
+        *   ``input_length`` is too large for the implementation.
```
athoelke, line 785: As per psa_sign_message()

doc/crypto/api/ops/signature.rst
```diff
@@ -681,6 +744,65 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The following conditions can result in this error:
+
+        *   ``alg`` is not an asymmetric signature algorithm that permits verifying a message with a context string.
+        *   ``key`` is not a public key or an asymmetric key pair, that is compatible with ``alg``.
+        *   ``input_length`` is too large for the algorithm and key type.
```
athoelke, line 791: As per psa_sign_message()

doc/crypto/api/ops/signature.rst
```diff
@@ -681,6 +744,65 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_COMMUNICATION_FAILURE
+    .. retval:: PSA_ERROR_CORRUPTION_DETECTED
+    .. retval:: PSA_ERROR_STORAGE_FAILURE
+    .. retval:: PSA_ERROR_DATA_CORRUPT
+    .. retval:: PSA_ERROR_DATA_INVALID
+    .. retval:: PSA_ERROR_BAD_STATE
+        The library requires initializing by a call to `psa_crypto_init()`.
+
+    ..  note::
+        To perform a multi-part hash-and-sign signature verification algorithm, first use a :ref:`multi-part hash operation <hash-mp>` to hash the message and then pass the resulting hash to `psa_verify_hash()`. :code:`PSA_ALG_GET_HASH(alg)` can be used to determine the hash algorithm to use.
```
athoelke, line 802: Again, refer to `psa_verify_hash_with_context()`

doc/crypto/api/ops/signature.rst
```diff
@@ -745,6 +867,75 @@ Asymmetric signature functions
 
     Specialized signature algorithms can apply a padding or encoding to the hash. In such cases, the encoded hash must be passed to this function. For example, see `PSA_ALG_RSA_PKCS1V15_SIGN_RAW`.
 
+.. function:: psa_sign_hash_with_context
+
+    .. summary::
+        Sign a pre-computed hash with a private key.
```
athoelke, line 873: Mention the non-default/non-empty context.

doc/crypto/api/ops/signature.rst
```diff
@@ -465,16 +465,13 @@ EdDSA signature algorithms
 
     *   Edwards25519: the Ed25519 algorithm is computed. The output signature is a 64-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.1.6`.
```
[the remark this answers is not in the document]
athoelke: I think that either way, we should not state as a general rule that `psa_sign_message()` is equivalent to `psa_sign_message_with_context()`:

* It doesn't work for Ed25519/Ed25519ctx, however we decide to present this
* I think I prefer that signature algorithms that do not have a context input result in a error when calling `psa_sign_message_with_context()`, even with an empty context parameter. But I can see an argument that applications might prefer to always call the _with_context function, and pass an empty context when one is not used by the algorithm?

doc/crypto/api/ops/signature.rst
```diff
@@ -745,6 +867,75 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_NOT_PERMITTED
+        The key does not have the `PSA_KEY_USAGE_SIGN_HASH` flag, or it does not permit the requested algorithm.
+    .. retval:: PSA_ERROR_BUFFER_TOO_SMALL
+        The size of the ``signature`` buffer is too small.
+        `PSA_SIGN_OUTPUT_SIZE()` or `PSA_SIGNATURE_MAX_SIZE` can be used to determine a sufficient buffer size.
+    .. retval:: PSA_ERROR_NOT_SUPPORTED
+        The following conditions can result in this error:
+
+        *   ``alg`` is not supported, or is not an asymmetric signature algorithm that permits signing a pre-computed hash.
+        *   ``key`` is not supported for use with ``alg``.
```
athoelke, line 916: items about the context, as per psa_sign_message().

doc/crypto/api/ops/signature.rst
```diff
@@ -745,6 +867,75 @@ Asymmetric signature functions
...
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The following conditions can result in this error:
+
+        *   ``alg`` is not an asymmetric signature algorithm that permits signing a pre-computed hash with a context.
+        *   ``key`` is not an asymmetric key pair, that is compatible with ``alg``.
+        *   ``hash_length`` is not valid for the algorithm and key type.
+        *   ``hash`` is not a valid input value for the algorithm and key type.
```
athoelke, line 923: items about the context, as per psa_sign_message().

doc/crypto/api/ops/signature.rst
```diff
@@ -801,6 +992,66 @@ Asymmetric signature functions
     Specialized signature algorithms can apply a padding or encoding to the hash. In such cases, the encoded hash must be passed to this function. For example, see `PSA_ALG_RSA_PKCS1V15_SIGN_RAW`.
 
 
+.. function:: psa_verify_hash_with_context
+
+    .. summary::
+        Verify the signature of a hash or short message using a public key.
```
athoelke, line 998: Mention context

doc/crypto/api/ops/signature.rst
```diff
@@ -801,6 +992,66 @@ Asymmetric signature functions
...
+
+        *   ``alg`` is not supported, or is not an asymmetric signature algorithm that permits verifying a pre-computed hash.
+        *   ``key`` is not supported for use with ``alg``.
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The following conditions can result in this error:
+
+        *   ``alg`` is not an asymmetric signature algorithm that permits verifying a pre-computed hash with a context.
+        *   ``key`` is not a public key or an asymmetric key pair, that is compatible with ``alg``.
+        *   ``hash_length`` is not valid for the algorithm and key type.
+        *   ``hash`` is not a valid input value for the algorithm and key type.
```
athoelke, line 1040: Items about the context parameter

doc/crypto/api/ops/signature.rst
```diff
@@ -801,6 +992,66 @@ Asymmetric signature functions
...
+        ``key`` is not a valid key identifier.
+    .. retval:: PSA_ERROR_NOT_PERMITTED
+        The key does not have the `PSA_KEY_USAGE_VERIFY_HASH` flag, or it does not permit the requested algorithm.
+    .. retval:: PSA_ERROR_INVALID_SIGNATURE
+        ``signature`` is not the result of signing ``hash`` with algorithm ``alg`` using the private key corresponding to ``key``.
+    .. retval:: PSA_ERROR_NOT_SUPPORTED
+        The following conditions can result in this error:
+
+        *   ``alg`` is not supported, or is not an asymmetric signature algorithm that permits verifying a pre-computed hash.
+        *   ``key`` is not supported for use with ``alg``.
```
athoelke, line 1033: Items about the context parameter

doc/crypto/api/ops/signature.rst
```diff
@@ -856,6 +1107,22 @@ Support macros
                          hash, sizeof(hash), &hash_len);
         psa_sign_hash(key, alg, hash, hash_len, ...);
 
+.. macro:: PSA_ALG_SUPPORTS_CONTEXT
```
athoelke, line 1110: Wondering about other possible naming options for this macro: `PSA_ALG_USES_CONTEXT()`...?

doc/crypto/api/ops/signature.rst
```diff
@@ -465,16 +465,13 @@ EdDSA signature algorithms
```
athoelke, line 465: The new functions should probably be added/mentioned in the "Signature functions" subsection in the chapter pre-amble.

athoelke (changes requested):
Thanks for the PR for this.

Overall:

* This will be a minor version addition (i.e. target 1.4.0)
* We need to decide how to handle the Ed25519ctx anomaly
* We need to decide whether it is valid to call a _with_context function with an empty context for an algorithm that does not accept contexts.

See the review comments for details. Aside from those, we should also add the new functions to the relevant usage flag descriptions, alongside the non-context signature functions.

doc/crypto/api/ops/signature.rst
MarcusJGStreets: Yes

doc/crypto/api/ops/signature.rst
MarcusJGStreets: I wondered about that, but settled on Supports. 
I have clarified the wording to note that it reflects the implementation of the algorithm.

doc/crypto/api/ops/signature.rst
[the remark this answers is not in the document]
MarcusJGStreets: I have added a new algorithm. 
I think that this is the best option.
It is an error to use PureEdDSA with a non-empty context - as with all algorithms that do not support contexts.
PureEdDSA always dies original, 25519

25519CTX always uses a context, but if you call psa_sign_message or if context_length is zero, it uses the empty string.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
```
athoelke, line 473: Three lines up (470): The algorithm can also be used with the sign/verify message with context functions. At least for Edwards448 curve keys.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
 
     PureEdDSA requires an elliptic curve key on a twisted Edwards curve. The following curves are supported:
 
-    *   Edwards25519: the Ed25519 algorithm is computed. The output signature is a 64-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.1.6`.
+    *   Edwards25519: the Ed25519 algorithm is computed. The output signature is a 64-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.1.6`.
```
athoelke, line 476: Might be worth stating that _with_context() is not supported, except with an zero-length context.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
 
     PureEdDSA requires an elliptic curve key on a twisted Edwards curve. The following curves are supported:
 
-    *   Edwards25519: the Ed25519 algorithm is computed. The output signature is a 64-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.1.6`.
+    *   Edwards25519: the Ed25519 algorithm is computed. The output signature is a 64-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.1.6`. 
 
-    *   Edwards448: Unless you use the `_with_context` functions, the Ed448 algorithm is computed with an empty string as the context. The output signature is a 114-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.2.6`.
+    *   Edwards448: Unless you use the signature functions that accept a context parameter, such as :code:`psa_sign_message_with_context()` and :code:`psa_verify_message_with_context()`, the Ed448 algorithm is computed with an empty string as the context. The output signature is a 114-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.2.6`.
```
athoelke, line 478: Should we try and be consistent with use of 'zero-length context' vs 'empty context' vs 'empty string as context'?

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
...
 
     .. note::
         To sign or verify the pre-computed hash of a message using EdDSA, the HashEdDSA algorithms (`PSA_ALG_ED25519PH` and `PSA_ALG_ED448PH`) can be used.
 
         The signature produced by HashEdDSA is distinct from that produced by PureEdDSA.
 
+    .. note::
+        When signatures on the Edwards 25519 curve were originally defined without domain separation. Later the Ed25519ctx and Ed25519ph variants were defined, both of which accept a context string. However, a signature made with Ed25519ctx and an empty context is distinct from a signature made using the Ed25519.
+        
+        As PureEdDSA does not support contexts, using PureEdDSA with a non-empty context is an error.
```
athoelke, line 488: This is only true for the 25519 curve. Ed448 is defined to take contexts, with domain separation - i.e `PSA_ALG_EDDSA` for Edwards448 keys works like `PSA_ALG_ED25519CTX` for Edwards25519 keys..

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
...
+    *   Edwards448: Unless you use the signature functions that accept a context parameter, such as :code:`psa_sign_message_with_context()` and :code:`psa_verify_message_with_context()`, the Ed448 algorithm is computed with an empty string as the context. The output signature is a 114-byte string: the concatenation of :math:`R` and :math:`S` as defined by :RFC:`8032#5.2.6`.
 
     .. note::
         To sign or verify the pre-computed hash of a message using EdDSA, the HashEdDSA algorithms (`PSA_ALG_ED25519PH` and `PSA_ALG_ED448PH`) can be used.
 
         The signature produced by HashEdDSA is distinct from that produced by PureEdDSA.
 
+    .. note::
+        When signatures on the Edwards 25519 curve were originally defined without domain separation. Later the Ed25519ctx and Ed25519ph variants were defined, both of which accept a context string. However, a signature made with Ed25519ctx and an empty context is distinct from a signature made using the Ed25519.
+
```
athoelke, line 487: In hindsight, I think we would have been better off without defining PSA_ALG_EDDSA at all, and had 5 separate algorithms for Edwards signatures (Ed25519, Ed25519ctx, Ed25519ph, Ed448, and Ed448ph). As it is we will now have four, and hard-to-document habaviour differences - PSA_ALG_SUPPORTS_CONTEXT(PSA_ALG_EDDSA) is indeterminate, as it depends on the key curve.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
...
+.. macro:: PSA_ALG_ED25519CTX
+    :definition: ((psa_algorithm_t) 0x0600090C)
+
+    .. summary::
+        Edwards-curve digital signature algorithm with context, using the Edwards25519 curve.
+
+        .. versionadded:: 1.4.0
+        
+    This signature algorithm can be used with both the message and hash signature functions.
```
athoelke, line 503: No, only with the message and message-with-context functions, as for PureEdDSA. Only the *ph variants can be used with the sign/verify_hash functions.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
...
+        
+        As PureEdDSA does not support contexts, using PureEdDSA with a non-empty context is an error. 
+
+    .. subsection:: Compatible key types
+
+        | :code:`PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_TWISTED_EDWARDS)`
+        | :code:`PSA_KEY_TYPE_ECC_PUBLIC_KEY(PSA_ECC_FAMILY_TWISTED_EDWARDS)` (signature verification only)
+
+.. macro:: PSA_ALG_ED25519CTX
+    :definition: ((psa_algorithm_t) 0x0600090C)
```
athoelke, line 496: Ed25519CTX probably needs a new SIGN-TYPE for the algorithm encoding. It would be nice if it was the same as Ed448, as they share the same formulation, aisde from the hash function. But too late for that.

Lets allocate 0x0a for the signature type. We could encode the hash algorithm (SHA512) in the ID, but there is no benefit to the application to know the implied hash function (as hash-and-sign is not supported). So my suggestion to keep it simple is `0x06000A00`.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
...
+        *   Call `psa_sign_message()` or `psa_sign_message_with_context()` with the message.
+
+        Verifying a signature is similar, using, for example, `psa_verify_message()` or `psa_verify_message_with_contex()` instead of the signature function.
+
     .. subsection:: Compatible key types
 
         | :code:`PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_TWISTED_EDWARDS)`
         | :code:`PSA_KEY_TYPE_ECC_PUBLIC_KEY(PSA_ECC_FAMILY_TWISTED_EDWARDS)` (signature verification only)
 
+    .. admonition:: Implementation note
```
athoelke, line 522: Move this above the Usage sub-section, as that is where the rendering will place content that is not inside its own subsection.

doc/crypto/api/ops/signature.rst
```diff
@@ -463,20 +473,56 @@ EdDSA signature algorithms
...
+        | :code:`PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_TWISTED_EDWARDS)`
+        | :code:`PSA_KEY_TYPE_ECC_PUBLIC_KEY(PSA_ECC_FAMILY_TWISTED_EDWARDS)` (signature verification only)
+
+.. macro:: PSA_ALG_ED25519CTX
+    :definition: ((psa_algorithm_t) 0x0600090C)
+
+    .. summary::
+        Edwards-curve digital signature algorithm with context, using the Edwards25519 curve.
+
+        .. versionadded:: 1.4.0
```
athoelke, line 501: Major.Minor only - so "1.4" here

athoelke:
We could possibly even go further, and combine the 'in hindsight option 1' with the alternative proposal:

* Introduce `PSA_ALG_ED25519` that implements Ed25519 (original form). This is equivalent to `PSA_ALG_EDDSA`, but only for 255-bit keys. This does not support a non-empty context.
* Introduce `PSA_ALG_EDDSA_CTX` that implements Ed25519ctx for 255-bit keys, and Ed448 for 448-bit keys. This algorithm supports a non-empty context.
* Specify that `PSA_ALG_EDDSA` cannot be used with a non-empty context. Deprecate the use of this algorithm identifier - it does not lend itself to algorithm agility. Clearly existing code and key stores using this algorithm should continue to work (without contexts).

doc/crypto/api/ops/signature.rst
MarcusJGStreets: Of course.

doc/crypto/api/ops/signature.rst
MarcusJGStreets: Yes, I will do so in the next commit

doc/crypto/api/ops/signature.rst
```diff
@@ -502,41 +504,73 @@ EdDSA signature algorithms
...
+        Verifying a signature is similar, using, for example, `psa_verify_message()` or `psa_verify_message_with_context()` instead of the signature function.
 
     .. subsection:: Compatible key types
 
         | :code:`PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_TWISTED_EDWARDS)`
         | :code:`PSA_KEY_TYPE_ECC_PUBLIC_KEY(PSA_ECC_FAMILY_TWISTED_EDWARDS)` (signature verification only)
 
 
-.. macro:: PSA_ALG_ED25519PH
+.. macro:: PSA_ALG_ED448CTX
```
athoelke, line 529: We want to enable algorithm/key-size agility by using a single algorithm identifier for "PureEdDSA with non-zero context": `PSA_ALG_EDDSA_CTX`. This identifier requires that the implementation perform Ed25519ctx with a 255-bit key and Ed448 with a 448-bit key.

This definition should be folded into the `PSA_ALG_ED25519CTX` and renamed to be key-size agnostic. We can use the 0x06000A00 value for `PSA_ALG_EDDSA_CTX`.

doc/crypto/api/ops/signature.rst
```diff
@@ -502,41 +504,73 @@ EdDSA signature algorithms
...
+        
+    This signature algorithm can be used with both the message and message with context  signature functions.
+
+    This calculates the Ed448ctx algorithm as specified in :RFC-title:`8032#5.1`, and requires an Edwards448 curve key. The `psa_sign_message()` and `psa_verify_message()` functions use an zero-length context when computing or verifying signatures. 
+    
+    To use a non-zero-length context, use the signature functions that accept a context parameter, such as :code:`psa_sign_message_with_context()` and :code:`psa_verify_message_with_context()`
+
+    .. admonition:: Implementation note
+
+       Even if you supply an zero-length context, signatures created with Ed448ctx are distinct from those created with PureEdDSA.
```
athoelke, line 545: Ed448ctx does not exist - this is a cut-n-paste error from the ED25519CTX algorithm. The discrepancy is between `PSA_ALG_PURE_EDDSA` and `PSA_ALG_EDDSA_CTX` for 255-bit keys.

athoelke (changes requested):
Needs further changes to combine `PSA_ALG_ED25519CTX` and `PSA_ALG_ED448CTX` into a single algorithm identier. Also see a response in the PR thread that summarises the expected outcome and behaviors.

doc/crypto/api/ops/signature.rst
```diff
@@ -56,6 +56,7 @@ There are three categories of asymmetric signature algorithm in the |API|:
     The following algorithms are in this category:
 
     | `PSA_ALG_PURE_EDDSA`
+      `PSA_ALG_EDDSA_CTX`
```
athoelke, line 59: This needs a leading '|' character to treat this as a paragraph of left-justified short lines

doc/crypto/api/ops/signature.rst
```diff
@@ -465,15 +476,48 @@ EdDSA signature algorithms
...
+    :definition: ((psa_algorithm_t) 0x06000A00)
+
+    .. summary::
+        Edwards-curve digital signature algorithm with context, using the Edwards25519 or Edwards448 curves.
+
+        .. versionadded:: 1.4
+        
+    This signature algorithm can be used with both the message and message with context signature functions. They cannot be used to sign hashes. 
+
+    This calculates the algorithm as specified in :RFC-title:`8032#5.1`, and requires an Edwards25519 or Edwards448 curve key. The `psa_sign_message()` and `psa_verify_message()` functions use an zero-length context when computing or verifying signatures.
```
athoelke, line 506: an → a

athoelke (approved):
I've fixed the issues that I identified as part of the copy-editor updates.

This would benefit from additional review to ensure I have not made other errors in that update.

athoelke:
We might need to split this into a PR for the main specification, and a separate PR for the PQC Extension.

PR #293, opened 2025-09-05
Add support for the WPA3-SAE PAKE (part of 802.11)

### LINKED ISSUE #203: Dragonfly PAKE for WPA3 use cases
WPA3, part of the IEEE 802.11 wireless specifications, defines a key establishment mechanism called *Simultaneous Authentication of Equals* (SAE). The key exchange mechanism is a variant of the Dragonfly PAKE (see [RFC 7664](https://www.rfc-editor.org/rfc/rfc7664.html)).

The SAE protocol has evolved:
* On first debut in 802.11s (2011) for mesh networks, SAE used a 'hunting-and-pecking' (HNP) method for computing a group element (ECC or FF) that is very similar to the one described in RFC 7664.
* SAE was adopted for WPA3-Personal in 802.11-2016, to replace the use of WPA2 (and WEP).
* Vulnerabilities in the WPA3-SAE protocol, particularly related to the password to group element derivation, and the scope for down-grade attacks, were published in 2019.
* Countermeasures are implemented in the 802.11-2020 specification, introducing a preferred 'hash-to-curve' (aka hash-to-element or H2E) method for the group element computation that can be implemented in constant time, and adding information to the key derivation context to mitigate a downgrade attack.

SAE is fully specified in IEE 802.11-2020 §12.4, including the H2E and HNP methods, the key exchange, and the specific hash and key derivation procedures for the protocol.

athoelke:
This is the full documentation PR for WPA3-SAE, building on the outline for the API in #230 contributed by Oberon.

It provides support for all cipher-suites and operational variants that are defined in IEEE-802.11-2024 §12.4:

* Password element derivation via *looping* or use of the *hash-to-element* procedure
* Fixed-size or group-dependent-hash-sized Key confirmation key (SAE-KCK) and Pairwise master key (PMK)
* Elliptic curve and finite field PAKE primitives

### Open items

- [x] 1. The password token key type names currently have a `_PT` suffix, following the contributed design. This is not required to distinguish the types, as there are no non-password-token key types for WPA3-SAE. I suggest removing them.
- [x] 2. ~~Done~~
- [x] 3. ~~Done~~
- [x] 4. The evolution of the WPA3-SAE protocol over 3 versions of IEEE-802.11 results in the password value being used in more than one type of algorithm:

   * If the looping method is used, the password is passed directly to the PAKE operation using `PSA_ALG_WPA3_SAE_FIXED(PSA_ALG_SHA256)`.
   * If the hash-to-element or group-dependent-hash variants are used, the password is used in a `PSA_ALG_WPA3_SAE_H2E()` KDF operation, parameterized by a hash related to the PAKE primitive in use.

   Either we will require the application (the 802.11 implementation of SAE) to store the password, and import a volatile password key with an appropriate policy once the variant and cipher suite is determined; or we would need a wildcard algorithm identifier for use in the password key policy.

   It might be sufficient to permit `PSA_ALG_ANY_HASH` to be used in `PSA_ALG_WPA3_SAE_H2E()` as a wildcard policy when using the password to derive the password token, and not provide support for looping vs hash-to-element use of a password key?

- [x] 5. A password token that is derived through the hash-to-element procedure is specific to the PAKE primitive (type of cyclic group, selected family, and size). The same password token can be used in both the `PSA_ALG_WPA3_SAE_FIXED()` and `PSA_ALG_WPA3_SAE_GDH()` algorithms - noting that the hash algorithm is effectively determined by the PAKE primitive.

   Is it valuable to have a wildcard permitted-algorithm policy for WPA3-SAE password token keys, so that the same key could be used in either the fixed-sized or group-dependent-hash-sized variants of WPA3-SAE?

Fixes #203

### DIFF
--- a/doc/crypto/api.db/psa/crypto.h
+++ b/doc/crypto/api.db/psa/crypto.h
@@ -190,7 +190,7 @@ typedef struct psa_custom_key_parameters_t {
 #define PSA_ALG_TLS12_PSK_TO_MS(hash_alg) /* specification-defined value */
 #define PSA_ALG_TRUNCATED_MAC(mac_alg, mac_length) \
     /* specification-defined value */
-#define PSA_ALG_WPA3_SAE_ANY ((psa_algorithm_t)0x0a000800)
+#define PSA_ALG_WPA3_SAE_ANY ((psa_algorithm_t)0x0a0088ff)
 #define PSA_ALG_WPA3_SAE_FIXED(hash_alg) /* specification-defined value */
 #define PSA_ALG_WPA3_SAE_GDH(hash_alg) /* specification-defined value */
 #define PSA_ALG_WPA3_SAE_H2E(hash_alg) /* specification-defined value */
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -37,7 +37,7 @@ The following algorithm policies are supported:
 *   An algorithm built from `PSA_ALG_AT_LEAST_THIS_LENGTH_MAC()` permits any MAC algorithm from the same base class (for example, CMAC) which computes or verifies a MAC length greater than or equal to the length encoded in the wildcard algorithm.
 *   An algorithm built from `PSA_ALG_AEAD_WITH_AT_LEAST_THIS_LENGTH_TAG()` permits any AEAD algorithm from the same base class (for example, CCM) which computes or verifies a tag length greater than or equal to the length encoded in the wildcard algorithm.
 *   The `PSA_ALG_CCM_STAR_ANY_TAG` wildcard algorithm permits the `PSA_ALG_CCM_STAR_NO_TAG` cipher algorithm, the `PSA_ALG_CCM` AEAD algorithm, and the :code:`PSA_ALG_AEAD_WITH_SHORTENED_TAG(PSA_ALG_CCM, tag_length)` truncated-tag AEAD algorithm for ``tag_length`` equal to 4, 8 or 16.
-*   The wildcard key policy :code:`PSA_ALG_WPA3_SAE_H2E(PSA_ALG_ANY_HASH)` permits a password key to be used with any WPA3-SAE cipher suite.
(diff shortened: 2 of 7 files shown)


athoelke:
I've update the PR to incorporate those changes (and updated the PR description to remove the related open issues).

We now need to conclude on the key type name and wildcard items.

oberon-sk:
> 1. The password token key type names currently have a `_PT` suffix, following the contributed design. This is not required to distinguish the types, as there are no non-password-token key types for WPA3-SAE. I suggest removing them.

Removing them should be fine.

> 4. The evolution of the WPA3-SAE protocol over 3 versions of IEEE-802.11 results in the password value being used in more than one type of algorithm:
>        
> * If the looping method is used, the password is passed directly to the PAKE operation using `PSA_ALG_WPA3_SAE_FIXED(PSA_ALG_SHA256)`.
> * If the hash-to-element or group-dependent-hash variants are used, the password is used in a `PSA_ALG_WPA3_SAE_H2E()` KDF operation, parameterized by a hash related to the PAKE primitive in use.
>        
> Either we will require the application (the 802.11 implementation of SAE) to store the password, and import a volatile password key with an appropriate policy once the variant and cipher suite is determined; or we would need a wildcard algorithm identifier for use in the password key policy.
> It might be sufficient to permit `PSA_ALG_ANY_HASH` to be used in `PSA_ALG_WPA3_SAE_H2E()` as a wildcard policy when using the password to derive the password token, and not provide support for looping vs hash-to-element use of a password key?

Yes, permitting `PSA_ALG_ANY_HASH` is sufficient.

> 5. A password token that is derived through the hash-to-element procedure is specific to the PAKE primitive (type of cyclic group, selected family, and size). The same password token can be used in both the `PSA_ALG_WPA3_SAE_FIXED()` and `PSA_ALG_WPA3_SAE_GDH()` algorithms - noting that the hash algorithm is effectively determined by the PAKE primitive.
> Is it valuable to have a wildcard permitted-algorithm policy for WPA3-SAE password token keys, so that the same key could be used in either the fixed-sized or group-dependent-hash-sized variants of WPA3-SAE?

Probably yes.

PR #297, opened 2025-09-12
Add support for the Ascon family of algorithms

### LINKED ISSUE #60: Ascon Algorithm Identifiers
ASCON-128, ASCON-128a, ASCON-HASH and ASCON-HASHA would fit into the PSA Certified API, with just the definition of some key types (for the AEAD) and algorithm Ids needed.

Are you going to define such constants for the PSA Certified API?

> Ascon is a family of [authenticated encryption](https://en.wikipedia.org/wiki/Authenticated_encryption) and [hashing](https://en.wikipedia.org/wiki/Cryptographic_hash_function) algorithms designed to be lightweight and easy to implement, even with added countermeasures against side-channel attacks. Ascon has been [selected as new standard](https://www.nist.gov/news-events/news/2023/02/nist-selects-lightweight-cryptography-algorithms-protect-small-devices) for lightweight cryptography in the [NIST Lightweight Cryptography competition (2019–2023)](https://csrc.nist.gov/projects/lightweight-cryptography/finalists). Ascon has also been selected as the primary choice for lightweight authenticated encryption in the final portfolio of the [CAESAR competition (2014–2019)](https://competitions.cr.yp.to/caesar-submissions.html).

(from https://ascon.iaik.tugraz.at/)

athoelke:
This adds support for all four algorithms defined in NIST SP800-232.

It uses the new API merged in #253.

Fixes #60

### DIFF
--- a/doc/crypto/api/ops/mac.rst
+++ b/doc/crypto/api/ops/mac.rst
@@ -55,20 +55,22 @@ MAC algorithms
 
     .. subsection:: Choice of hash algorithm
 
-        HMAC was designed with hashes that use a Merkle-Damgård construction, for example, MD5, SHA-1, and SHA-2.
-        In particular, the input-block size of the hash must be at least as large as the output size.
+        An HMAC block size must be defined for use with each hash algorithm, which is at least as large as the hash output size.
 
-        Hash algorithms which have a small input-block size are not directly suitable for use with HMAC.
+        HMAC was designed for hashes that use a Merkle-Damgård construction, for example, MD5, SHA-1, and SHA-2.
+        For these hash algorithms, the HMAC block size is defined to be the hash input-block size.
+
+        Some algorithms do not have a defined HMAC block size.
         For example, Ascon (`PSA_ALG_ASCON_HASH256`) or Shake-based hashes (`PSA_ALG_SHAKE256_512`).
 
-        :numref:`tab-hmac-hash` lists the valid hash algorithms for use with HMAC, and their input and output byte-sizes.
+        :numref:`tab-hmac-hash` lists the valid hash algorithms for use with HMAC, and their HMAC block and output sizes in bytes.
 
         .. csv-table:: Hash algorithms that can be used with HMAC
             :name: tab-hmac-hash
             :header-rows: 1
             :widths: 3 2 2
 
-            Algorithm, Input block size, Output size
+            Algorithm, HMAC block size, Output size
             `PSA_ALG_MD2`, 16, 16
             `PSA_ALG_MD4`, 64, 16
             `PSA_ALG_MD5`, 64, 16
@@ -88,8 +90,8 @@ MAC algorithms
 
         .. admonition:: Implementation note
 
-            It is recommended that short-block-length hash algorithms are not supported with `PSA_ALG_HMAC`.
-            Future versions of the |API| might specify support for these hash algorithms, and will define the input block size to use for each hash algorithm.
+            It is recommended that other hash algorithms are not supported with `PSA_ALG_HMAC`.
+            Future versions of the |API| might specify HMAC support for these hash algorithms, and will define the block size to use for HMAC.
 
     .. subsection:: Compatible key types
 


doc/crypto/api/ops/hash.rst
```diff
@@ -206,6 +206,19 @@ Hash algorithms
 
     SM3 is defined in :cite-title:`ISO10118`, and also in :cite-title:`CSTC0004`.
 
+.. macro:: PSA_ALG_ASCON_HASH256
+    :definition: ((psa_algorithm_t)0x02000020)
+
+    .. summary::
+        The Ascon-Hash256 message-digest algorithm.
```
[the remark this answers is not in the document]
gilles-peskine-arm: We don't follow any particular national standard. NIST has never approved anything using MD5, but there's no ambiguity about what HMAC-MD5 means. The ambiguity for Ascon is that I'm not sure if there's a consensus about the block size, and in any case, as a reader of the specification, I wouldn't know where to look.

MD[245], RIPEMD-160 and SM3 is a Merkle-Damgård construction so the traditional definition of HMAC specifies the block length unambiguously. For SHA3, there was a wide consensus for the block length for HMAC even before NIST endorsed it.

For the SHAKE-based hashes that are only intended for use inside specific higher-level constructions, I doubt anyone cares about HMAC. But for Ascon, it's a reasonable concern. That explains why we hadn't noticed this omission before.

I think we should do one of the following for HMAC, with no strong preference:

* Ascon and SHAKExxx: forbidden for now, but future versions of the specification may allow it and will specify the block size then.
* Ascon and SHAKExxx: reserved, and implementations may implement it but are warned that a future version of the specification may allow it and will specify the block size then.
* Ascon reserved, SHAKExxx forbidden.
* Ascon = 64 (with a citation or more — I haven't investigated how widespread the choice of 64 is), SHAKExxx reserved.
* Ascon = 64, SHAKExxx forbidden.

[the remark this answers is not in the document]
athoelke: I've added a section to HMAC to describe the choice of hash algorithm.

doc/crypto/api/ops/mac.rst
```diff
@@ -49,6 +51,46 @@ MAC algorithms
 
     The HMAC construction is defined in :RFC-title:`2104`.
 
+    .. _hash-for-hmac:
+
+    .. subsection:: Choice of hash algorithm
+
+        HMAC was designed with hashes that use a Merkle-Damgård construction, for example, MD5, SHA-1, and SHA-2.
```
gilles-peskine-arm: ITYM “was designed with hashes ... in mind” or “was designed for hashes”.

doc/crypto/api/ops/mac.rst
```diff
@@ -49,6 +51,46 @@ MAC algorithms
...
+        In particular, the input-block size of the hash must be at least as large as the output size.
+
+        Hash algorithms which have a small input-block size are not directly suitable for use with HMAC.
+        For example, Ascon (`PSA_ALG_ASCON_HASH256`) or Shake-based hashes (`PSA_ALG_SHAKE256_512`).
```
gilles-peskine-arm: [SHAKE](https://en.wikipedia.org/wiki/SHA-3#Instances) does not have a small block size. Quite the opposite, the rate (which is the input-block size for the sponge construction that SHA3 and SHAKE are based on) is larger than for most hashes.

I'm not aware of a technical reason for not using HMAC-SHAKE128 or HMAC-SHAKE256 with the HMAC block size equal to the rate. However, as far as I know, using SHAKE as a hash is something that people only do inside specific algorithms, and I don't know of a good reason not to use one of the actual SHA3 variants if you want HMAC. (Or use KMAC if you want a MAC that's close to SHAKE.)

doc/crypto/api/ops/mac.rst
```diff
@@ -49,6 +51,46 @@ MAC algorithms
...
+            `PSA_ALG_SHA_512_256`, 128, 32
+            `PSA_ALG_SHA3_224`, 144, 28
+            `PSA_ALG_SHA3_256`, 136, 32
+            `PSA_ALG_SHA3_384`, 104, 48
+            `PSA_ALG_SHA3_512`, 72, 64
+            `PSA_ALG_SM3`, 64, 32
+
+        .. admonition:: Implementation note
+
+            It is recommended that short-block-length hash algorithms are not supported with `PSA_ALG_HMAC`.
```
gilles-peskine-arm: As discussed above, it's not so much about “short-block-length hash algorithms” as about hash algorithms for which there is no consensus on what block size to use for HMAC.

doc/crypto/api/ops/mac.rst
```diff
@@ -49,6 +51,46 @@ MAC algorithms
...
+            Future versions of the |API| might specify support for these hash algorithms, and will define the input block size to use for each hash algorithm.
```
gilles-peskine-arm: As discussed above, not the “input block size”, but the block size for HMAC.

doc/crypto/api/ops/mac.rst
```diff
@@ -49,6 +51,46 @@ MAC algorithms
 
     The HMAC construction is defined in :RFC-title:`2104`.
 
+    .. _hash-for-hmac:
+
+    .. subsection:: Choice of hash algorithm
+
+        HMAC was designed with hashes that use a Merkle-Damgård construction, for example, MD5, SHA-1, and SHA-2.
+        In particular, the input-block size of the hash must be at least as large as the output size.
```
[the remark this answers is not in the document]
athoelke: I'll change the wording to 'HMAC block size'

PR #312, opened 2025-10-15
Relax the permitted-algorithm policy for ECDSA

### LINKED ISSUE #278: Deterministic ML-DSA/SLH-DSA key policies not consistent with ECDSA
Reported by an implementation developer:

> Yesterday, another API question regarding ML-DSA came up. The beta spec requires that a key policy that permits PSA_ALG_ML_DSA should be combatible with PSA_ALG_DETERMINISTIC_ML_DSA. This is limited to psa_verify_message() for obvious reasons.
 >
> To be consistent, this special rule should be also applied to PSA_ALG_ECDSA/PSA_ALG_DETERMINISTIC_ECDSA. However, wouldn't it be simpler to remove the special rule for ML-DSA altogether? We think the benefit of that rule is rather small.

athoelke:
Fixes #278

### DIFF
--- a/doc/crypto/api/keys/policy.rst
+++ b/doc/crypto/api/keys/policy.rst
@@ -26,20 +26,26 @@ Permitted algorithms
 
 The permitted algorithm is encoded using a algorithm identifier, as described in :secref:`algorithms`.
 
-This specification only defines policies that restrict keys to a single algorithm, which is consistent with both common practice and security good practice.
+For most algorithms, this specification only defines policies that restrict keys to a single algorithm, which is consistent with both common practice and security good practice.
 
-The following algorithm policies are supported:
+If the permitted algorithm is `PSA_ALG_NONE`, no cryptographic operation is permitted with the key.
+The key can still be used for non-cryptographic actions such as exporting, if permitted by the usage flags.
 
-*   `PSA_ALG_NONE` does not permit any cryptographic operation with the key. The key can still be used for non-cryptographic actions such as exporting, if permitted by the usage flags.
-*   A specific algorithm value permits exactly that particular algorithm.
-*   A signature algorithm constructed with `PSA_ALG_ANY_HASH` permits the specified signature scheme with any hash algorithm. In addition, :code:`PSA_ALG_RSA_PKCS1V15_SIGN(PSA_ALG_ANY_HASH)` also permits the `PSA_ALG_RSA_PKCS1V15_SIGN_RAW` signature algorithm.
+For a cryptographic operation, the permitted algorithm value must exactly match the requested algorithm, except in the following cases:
+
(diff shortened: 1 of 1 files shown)


doc/crypto/appendix/history.rst
```diff
@@ -39,6 +39,12 @@ Changes to the API
 
     See :secref:`pake-wpa3-sae`.
 
+Relaxations
+~~~~~~~~~~~
+
+*   Relaxed the permitted-key policy requirements for ECDSA verification, to be consistent with those for ML-DSA and SLH-DSA.
```
gilles-peskine-arm: I think these exceptions should be mentioned in the “Permitted algorithms” section of the chapter on policies. Not necessarily the full list, but at least the general ideas that some algorithms are considered partially equivalent. Maybe just say that it's the case for some signature algorithms, which are only considered equivalent for verification? According to the current text, “A specific algorithm value permits exactly that particular algorithm” (plus NONE and various wildcard cases), which rules out the ML-DSA, SLH-DSA and now ECDSA verification relaxation.

gilles-peskine-arm (changes requested):
LGTM for the new text, but I think the section on policies also needs to be adapted.

doc/crypto/appendix/history.rst
athoelke: Good point - the PQC extension didn't actually modify that text, because it is outside the main spec for now.

doc/crypto/api/ops/signature.rst
```diff
@@ -334,15 +334,20 @@ ECDSA signature algorithms
...
 
-    This signature scheme is defined by :cite-title:`SEC1`, and also by :cite-title:`X9-62`, with a random per-message secret number :math:`k`.
+    .. note::
 
-    The representation of the signature as a byte string consists of the concatenation of the signature values :math:`r` and :math:`s`. Each of :math:`r` and :math:`s` is encoded as an :math:`N`-octet string, where :math:`N` is the length of the base point of the curve in octets. Each value is represented in big-endian order, with the most significant octet first.
+        It is impossible to determine whether a signature was produced with deterministic ECDSA or with randomized ECDSA: it is only possible to verify that a signature was made with ECDSA with the private key corresponding to the public key used for the verification.
```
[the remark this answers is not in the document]
mpg: Honestly that's probably just me being too picky here, and I'm not insisting on fixing this. But perhaps we could avoid the word "impossible" and just say "Signature verification can't determine ..."?

mpg (approved):

MarcusJGStreets (approved):
I am happy the intro includes the relaxation - which I also think is sensible. 
So I think this answers Giles point. 
Therefore, I am happy

PR #320, opened 2025-11-24
Update link to PSA-FFM

nicola-mazzucato-arm:
Current URL for PSA-FFM returns "We could not find this page".
Update it with the correct for Arm PSA Firmware Framework 1.0.


athoelke:
Thank you.

Somehow that reference was changed with all the other specifications two years ago, but somehow this one had an extra `.html` suffis which makes it useless.

PR #329, opened 2026-01-22
Publish Crypto 1.4.1, PQC Extension Final-1, and Status Code 1.0.5

athoelke:
Patch updates to specifications


site-src/crypto/1.4/about/about.html
```diff
@@ -205,12 +205,17 @@ <h3 id="searchlabel">Quick search</h3>
 <p>Added key query and key registration functions.</p>
 </td>
 </tr>
+<tr class="row-even"><td><p>January 2026</p></td>
+<td><p>1.4.1 Final</p></td>
+<td><p>Non-confidential</p></td>
+<td><p>Clarified hash algorithm usage in compound algorithms.</p></td>
```
gilles-peskine-arm: We've added OID for RSA. I thought we were also going to discuss OID for ML-DSA and SLH-DSA in the 1.4 PQC extension?

athoelke:
I've now added a patch release for the PQC Extension to include hash algorithm OID tables for HashML-DSA and HashSLH-DSA.

PR #331, opened 2026-01-23
Add table of OIDs for use when implementing HashML-DSA and HashSLH-DSA

### LINKED ISSUE #323: Missing hash algorithm for HashML-DSA
As discussed in our call today, FIPS 204 specifies that HashML-DSA may only be used with one of three hash functions: SHA-256, SHA-512, or SHAKE128/256. In order to be able to use these hash functions with `psa_sign_hash` and `psa_verify_hash`, we need a hash algorithm identifier for them. We're missing SHAKE128/256. (We have SHAKE256/512 for a similar reason, namely that it's used for Ed448.)

Also consider listing the OID involved in HashML-DSA, similar to the table that https://github.com/ARM-software/psa-api/pull/322 adds for RSA-PKCS#1v1.5. I don't know if we can expect all similar algorithms to use the same OIDs. (Certainly there are different OIDs for MD5, SHA1, SHA-256, etc., in different contexts, but until version 1.4 of the API specification, the only place where they appeared in our specification was RSA PKCS#1v1.5.)

Also check if SLH-DSA (FIPS 205) needs something similar.

athoelke:
Fixes #323

### DIFF
--- a/doc/ext-pqc/api/mldsa.rst
+++ b/doc/ext-pqc/api/mldsa.rst
@@ -182,24 +182,25 @@ An ML-DSA signature can only be verified with an ML-DSA algorithm.
 A HashML-DSA signature can only be verified with a HashML-DSA algorithm.
 
 :numref:`tab-ml-dsa-oid` lists the hash algorithm OIDs to use with the HashML-DSA algorithm.
+Note that for HashML-DSA the DER-encoded OID includes the tag and length.
 
 .. csv-table:: Hash algorithm OID to use in HashML-DSA
     :name: tab-ml-dsa-oid
     :header-rows: 1
     :class: longtable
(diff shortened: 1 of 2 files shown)


athoelke:
FIPS 204 and FIPS 205 state that the hash algorithm identifier string used in the Hash variants must have the OID tag and length included. I will amend this PR to add those bytes to the hex strings here.

For RSASSA-PKCS#1, the OID is enclosed in an AlgorithmIdentifier structure, that must also have an explicit NULL parameters field (after the OID); which is itself included in the DigestInfoValue structure that has the hash value string following the AlgorithmIdentifier. We could clarify that the hex values provided in the table for RSASSA-PKCS#1 are just the OID data values for each hash algorithm?

PR #335, opened 2026-01-28
Add multi-part operations for asymmetric signature

### LINKED ISSUE #319: Allow external mu to MLDSA sign/verify PSA APIs
In [NIST MLDSA specs](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf) supplying external mu is allowed instead of computing it internally in the API.
`Algorithm 8 ML-DSA.Verify_internal(𝑝𝑘, 𝑀 ′, 𝜎)
.
.
.
7: 𝜇 ← (H(BytesToBits(𝑡𝑟)||𝑀 ′, 64)) ▷ message representative that may optionally be
computed in a different cryptographic module`

This is useful in base data the needs to be signed/verified is scattered so supplying it in one continuous memory location can be a problem on memory constrained devices

athoelke:
This adds the new operations for multipart sign and multi-part verify for asymmetric signatures (previous discussions suggested these should be ditinct types).

* Updated the Functionality chapter to indicate the availability of multi-part operations for signature (and added two missed additions in 1.4).
* Extended the introduction to Asymmetric Signatures to describe the multi-part operations, and how they extend multi-part capability to message-signature algorithms.
* Added/modified text in signature algorithms to reflect their compatibility with the mluti-part operations.
* Modified the Notes in the single-part functions which recommended the use of a multi-part hash operation for fragmented messages.
* Add new operation types and functions.

Fixes #319

### DIFF
--- a/doc/crypto/api.db/psa/crypto.h
+++ b/doc/crypto/api.db/psa/crypto.h
@@ -861,11 +861,12 @@ psa_status_t psa_sign_message_with_context(psa_key_id_t key,
                                            size_t signature_size,
                                            size_t * signature_length);
 psa_sign_operation_t psa_sign_operation_init(void);
+psa_status_t psa_sign_set_context(psa_sign_operation_t * operation,
+                                  const uint8_t * context,
+                                  size_t context_length);
 psa_status_t psa_sign_setup(psa_sign_operation_t * operation,
                             psa_key_id_t key,
-                            psa_algorithm_t alg,
-                            const uint8_t * context,
-                            size_t context_length);
+                            psa_algorithm_t alg);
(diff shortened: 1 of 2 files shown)


athoelke:
This needs to wait for the 1.4.1 publication to be finalised.

One aspect of the proposed API that might be debatable:

* I've added context parameters to the `setup` functions, which can be zero-length (as per the `_with_context` single part functions). This seems to be a better approach than adding an additional, optional to call, `xxx_set_context()`function to each operation - which adds additional states and complexity to the implementation.

MarcusJGStreets (commented):
Looks to be the obvious way to implement multipart sign and verify.

PR #375, opened 2026-07-24
Add interruptible Crypto API operations

### LINKED ISSUE #23: Interruptible/bounded-latency asymmetric operations

### LINKED ISSUE #198: Support for interruptible key agreement/exchange

athoelke:
This is a rebased version of #107 and #199, targetting version 1.6 of the specification.

Apart from adding the previously drafted APIs, this incorporates other changes to the documentation since the original PRs. The introductory text for interruptible signatures has also been updated to reflect the presence of both multi-part signatures and interruptibel signature APIs.


doc/crypto/api/keys/management.rst
```diff
@@ -237,6 +237,9 @@ When creating a key, the attributes for the new key are specified in a `psa_key_
 
         This function is equivalent to calling `psa_generate_key_custom()` with the production parameters `PSA_CUSTOM_KEY_PARAMETERS_INIT` and ``custom_data_length == 0`` (``custom_data`` is ignored).
 
+    If an application requires bounded execution when generating a key, the implementation might provide support for interruptible key generation.
```
gilles-peskine-arm:
```suggestion
    If an application requires bounded execution time when generating a key, the implementation might provide support for interruptible key generation.
```

(“Bounded execution” is not a term I'm familiar with, and it seems to have a new meaning for AI agents.)

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+
+Interruptible key generation
+----------------------------
+
+Generation of some key types can be computationally expensive.
+For example, RSA keys, and elliptic curve public keys.
+
+For such keys, an interruptible key-generation operation can be used instead of calling `psa_generate_key()`, in applications that have bounded execution requirements for use cases that require key generation.
+
+This operation uses the default production parameters of `psa_generate_key()`. It does not support the custom production parameters accepted by `psa_generate_key_custom()`.
```
gilles-peskine-arm: How about adding a function `psa_generate_key_iop_custom()` to be optionally called after setup?

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+        The modulus is a product of two probabilistic primes between :math:`2^{n-1}` and :math:`2^n` where :math:`n` is the bit size specified in the attributes.
+
+    After a successful call to `psa_generate_key_iop_setup()`, the operation is active.
+    The operation can be completed by calling `psa_generate_key_iop_complete()` repeatedly, until it returns a status code that is not :code:`PSA_OPERATION_INCOMPLETE`.
+    Once active, the application must eventually terminate the operation. The following events terminate an operation:
+
+    *   A successful call to `psa_generate_key_iop_complete()`.
+    *   A call to `psa_generate_key_iop_abort()`.
+
+    If `psa_generate_key_iop_setup()` returns an error, the operation object is unchanged.
```
gilles-peskine-arm: This is not quite true for interruptible operation objects: a failed setup may reset the op count to 0.

(At least, I see no reason to forbid that, and it would be a burden to preserve the op count on all failure paths.)

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+
+        .. versionadded:: 1.6
+
+    .. param:: psa_generate_key_iop_t * operation
+        The interruptible key-generation operation to inspect.
+
+    .. return:: uint32_t
+        Number of *ops* that the operation has taken so far.
+
+    After the interruptible operation has completed, the returned value is the number of *ops* required for the entire operation.
```
gilles-peskine-arm: “required” makes me think this is about what has not been done yet. How about “the number of _ops_ spent on the entire operation”?

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+    The value is reset to zero by a call to either `psa_generate_key_iop_setup()` or `psa_generate_key_iop_abort()`.
```
gilles-peskine-arm: I don't think an unsuccessful call to `xxx_iop_setup()` should be guaranteed to reset the op count. In particular I think this is a valid implementation structure:
```
xxx_iop_setup(xxx_iop_t *operation) {
    if (operation->active) return PSA_ERROR_BAD_STATE;
    operation->ops = 0;
    status = xxx_iop_setup_internal(operation);
    if (status == PSA_SUCCESS) operation->active = 1; else xxx_op_abort(operation);
    return status;
}
```

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+        .. note::
+            This is an input parameter: it is not updated with the final key attributes.
+            The final attributes of the new key can be queried by calling `psa_get_key_attributes()` with the key's identifier.
+
+    .. return:: psa_status_t
+    .. retval:: PSA_SUCCESS
+        Success.
+        The interruptible operation must now be completed by calling `psa_generate_key_iop_complete()`.
+    .. retval:: PSA_ERROR_ALREADY_EXISTS
+        This is an attempt to create a persistent key, and there is already a persistent key with the given identifier.
```
gilles-peskine-arm, line 988: It may be worth noting that it's unspecified whether this error is detected at setup or complete time.

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+
+        .. note::
+            This is an input parameter: it is not updated with the final key attributes.
+            The final attributes of the new key can be queried by calling `psa_get_key_attributes()` with the key's identifier.
+
+    .. return:: psa_status_t
+    .. retval:: PSA_SUCCESS
+        Success.
+        The interruptible operation must now be completed by calling `psa_generate_key_iop_complete()`.
+    .. retval:: PSA_ERROR_ALREADY_EXISTS
```
gilles-peskine-arm, line 987: An implementation may or may not have an observable state of a persistent key identifier where this key id is reserved (so attempting to create a key with that id fails), but the key id doesn't exist (so e.g. `psa_get_key_attributes` returns `PSA_ERROR_INVALID_HANDLE`). Before interruptible key creation operations, this could happen in multithreaded environments. Now this can also happen if the implementation reserves the key identifier during an iop setup. I don't think the spec discusses this anywhere. Should it?

doc/crypto/api/ops/key-agreement.rst
```diff
@@ -352,6 +358,290 @@ Combining key agreement and key derivation
...
+        Attempt to finish a key agreement and return the shared secret.
+
+        .. versionadded:: 1.6
+
+    .. param:: psa_key_agreement_iop_t * operation
+        The interruptible key-agreement operation to use.
+        The operation must be active.
+    .. param:: psa_key_id_t * key
+        On success, an identifier for the newly created key.
+        `PSA_KEY_ID_NULL` on failure.
```
gilles-peskine-arm: Or if the operation is incomplete (which is neither really a success nor really a failure).

doc/crypto/api/ops/signature.rst
```diff
@@ -1244,6 +1268,7 @@ Context values are arbitrary strings between zero and 255 bytes in length.
 
 *   The signature functions without a context parameter provide a zero-length context when computing or verifying ML-DSA signatures.
 *   To provide a context, use the ``psa_xxxx_with_context()`` signature functions with a context parameter, such as :code:`psa_sign_message_with_context()`.
+*   Interruptible signature operations use a zero-length context.
```
gilles-peskine-arm: Why not have a `set_context` function for interruptible operations?

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+An interruptible asymmetric signature operation is used as follows:
+
+1.  Allocate an interruptible asymmetric signature operation object, of type `psa_sign_iop_t`, which will be passed to all the functions listed here.
+#.  Initialize the operation object with one of the methods described in the documentation for `psa_sign_iop_t`, for example, `PSA_SIGN_IOP_INIT`.
+#.  Call `psa_sign_iop_setup()` to specify the algorithm and key.
+#.  Call `psa_sign_iop_setup_complete()` to complete the setup, until this function does not return :code:`PSA_OPERATION_INCOMPLETE`.
+#.  Either:
+
+    1.  Call `psa_sign_iop_hash()` with a pre-computed hash of the message to sign; or
+    2.  Call `psa_sign_iop_update()` one or more times, passing a fragment of the message each time. The signature that is calculated will that be of the concatenation of these fragments, in order.
```
gilles-peskine-arm: Technically, zero or more times, since the message can be empty.

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+.. note::
+    An implementation of the |API| does not need to provide incremental generation for all key types supported by the implementation.
+    Use `psa_generate_key()` to create keys for types that do not need to be incrementally generated.
+
+An interruptible key-generation operation is used as follows:
+
+1.  Allocate an interruptible key-generation operation object, of type `psa_generate_key_iop_t`, which will be passed to all the functions listed here.
+#.  Initialize the operation object with one of the methods described in the documentation for `psa_generate_key_iop_t`, for example, `PSA_GENERATE_KEY_IOP_INIT`.
+#.  Call `psa_generate_key_iop_setup()` to specify the key attributes.
+#.  Call `psa_generate_key_iop_complete()` to finish generating the key, until this function does not return :code:`PSA_OPERATION_INCOMPLETE`.
```
gilles-peskine-arm: I find “does not return `PSA_OPERATION_INCOMPLETE`” weird: if the function is still running, it's not returning `PSA_OPERATION_INCOMPLETE` or anything else. I'd prefer “until this function returns a value other than `PSA_OPERATION_INCOMPLETE`”, even if it's longer.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+        *   The operation state is not valid: the operation must be set up, with no data input.
+        *   The library requires initializing by a call to `psa_crypto_init()`.
+    .. retval:: PSA_ERROR_NOT_PERMITTED
+        The key does not have the `PSA_KEY_USAGE_SIGN_HASH` flag.
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The following conditions can result in this error:
+
+        *   The algorithm does not allow signing of a pre-computed hash.
+        *   ``hash_length`` is not valid for the algorithm and key type.
+        *   ``hash`` is not a valid input value for the algorithm and key type.
```
gilles-peskine-arm, line 3074: Is there a way that `hash` can be invalid that isn't about `hash_length`?

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+    .. param:: size_t input_length
+        Size of the ``input`` buffer in bytes.
+
+    .. return:: psa_status_t
+    .. retval:: PSA_SUCCESS
+        Success.
+    .. retval:: PSA_ERROR_BAD_STATE
+        The following conditions can result in this error:
+
+        *   The operation state is not valid: the operation must be set up, with no pre-computed hash value input.
```
gilles-peskine-arm: And completion must not have started.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+    .. retval:: PSA_ERROR_COMMUNICATION_FAILURE
+    .. retval:: PSA_ERROR_CORRUPTION_DETECTED
+    .. retval:: PSA_ERROR_STORAGE_FAILURE
+    .. retval:: PSA_ERROR_DATA_CORRUPT
+    .. retval:: PSA_ERROR_DATA_INVALID
+    .. retval:: PSA_ERROR_INSUFFICIENT_ENTROPY
+
+    The application must complete the setup of the operation before calling this function.
+
+    For message-signature algorithms that process the message data multiple times when computing a signature, `psa_sign_iop_update()` must be called exactly once with the entire message content. For signature algorithms that only process the message data once, the message content can be passed in a series of calls to `psa_sign_iop_update()`.
```
gilles-peskine-arm, line 3141: Which error code should be used on a second `update()` to signal that the algorithm or implementation only allows a single update?

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,493 @@ Key export
...
+
+    *   A successful call to `psa_generate_key_iop_complete()`.
+    *   A call to `psa_generate_key_iop_abort()`.
+
+    If `psa_generate_key_iop_setup()` returns an error, the operation object is unchanged.
+
+.. function:: psa_generate_key_iop_complete
+
+    .. summary::
+        Attempt to finish the interruptible generation of a key.
```
gilles-peskine-arm: I don't think “Attempt to” is the right word here. I understand that as either succeeding, or doing nothing. But the expectation is that each call to `complete()` makes some progress towards the goal. So I think the description should be something like “Progress the interruptible …”.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+
+    After input of the message, the signature operation can be completed by calling `psa_sign_iop_complete()` until it returns a status code that is not :code:`PSA_OPERATION_INCOMPLETE`.
+
+    If this function returns an error status, the operation enters an error state and must be aborted by calling `psa_sign_iop_abort()`.
+
+    .. note::
+
+        To sign the zero-length message using an interruptible operation, call `psa_sign_iop_update()` once with a zero-length message fragment before calling `psa_sign_iop_complete()`.
```
gilles-peskine-arm: I think this note belongs on the complete() function instead of, or in addition to, here. Technically, if you meant to specifically sign an empty message, you might not think that you should read the documentation of the update function. Though in practice, a zero-length message is very likely to be the result of a fragmentation mechanism where the message happens to be empty at runtime.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+    .. return:: psa_status_t
+    .. retval:: PSA_SUCCESS
+        Success.
+        The first ``(*signature_length)`` bytes of ``signature`` contain the signature value.
+    .. retval:: PSA_OPERATION_INCOMPLETE
+        The function was interrupted after exhausting the maximum *ops*. The computation is incomplete, and this function must be called again with the same operation object to continue.
+    .. retval:: PSA_ERROR_BAD_STATE
+        The following conditions can result in this error:
+
+        *   The operation state is not valid: the operation setup must be complete, or a previous call to `psa_sign_iop_complete()` returned :code:`PSA_OPERATION_INCOMPLETE`.
```
gilles-peskine-arm: The first `BAD_STATE` condition on `psa_sign_iop_complete()` and `psa_verify_iop_complete()` seems wrong. The actual condition is that the operation object must be active and must have had input.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,14 +2783,752 @@ Multi-part asymmetric signature operations
...
+
+        .. versionadded:: 1.6
+
+    .. param:: psa_verify_iop_t * operation
+        The interruptible verification operation to set up. It must have been initialized as per the documentation for `psa_verify_iop_t` and not yet in use.
+    .. param:: psa_key_id_t key
+        Identifier of the key to use for the operation. It must be an asymmetric key pair or asymmetric public key. The key must either permit the usage `PSA_KEY_USAGE_VERIFY_HASH` or `PSA_KEY_USAGE_VERIFY_MESSAGE`.
+    .. param:: psa_algorithm_t alg
+        An asymmetric signature algorithm: a value of type `psa_algorithm_t` such that :code:`PSA_ALG_IS_SIGN(alg)` is true.
+    .. param:: const uint8_t * signature
```
gilles-peskine-arm: Can you remind me why the signature is passed during setup, rather than after the message?

Streaming protocols often send the message before the signature to verify, so in practice, this is likely to be a burden if the algorithm doesn't follow the sign-the-hash paradigm.

doc/crypto/figure/interruptible_operation_complex.puml
```diff
@@ -0,0 +1,44 @@
...
+
+[*] --> inactive: **Initialize**
+note as N1
+   Operation object starts as
+   uninitialised memory
+end note
+inactive --> setup: **Setup**\n//success//
+setup --> setup: **Setup-complete**\n//incomplete//
+setup --> input: **Setup-complete**\n//success//
+input --> input: **Update**\n//success//
```
gilles-peskine-arm, line 30: This doesn't take the hash-vs-message choice for sign/verify. I'm not sure if it should.

doc/crypto/api.db/psa/crypto.h
```diff
@@ -852,6 +892,23 @@ psa_status_t psa_sign_hash_with_context(psa_key_id_t key,
...
+                               const uint8_t * hash,
+                               size_t hash_length);
+psa_sign_iop_t psa_sign_iop_init(void);
+psa_status_t psa_sign_iop_setup(psa_sign_iop_t * operation,
+                                psa_key_id_t key,
+                                psa_algorithm_t alg);
+psa_status_t psa_sign_iop_setup_complete(psa_sign_iop_t * operation);
+psa_status_t psa_sign_iop_update(psa_sign_iop_t * operation,
+                                 const uint8_t * input,
+                                 size_t input_length);
```
gilles-peskine-arm, line 920: Coming back to the naming discussion (https://github.com/GlobalPlatform/psa-api/discussions/98, https://github.com/GlobalPlatform/psa-api/issues/23), there are a couple of things I don't like with the names here. `setup()` is used both for the first stage of a multipart operation and for the first stage of an interruptible operation. This doesn't help clear confusions between interruptible and multipart.

It also means that you can't tell the operation flow from looking at a function's name, specifically in the case of `xxx_iop_setup()`: that doesn't tell you whether there's a companion `xxx_iop_setup_complete()` function.

I think it would be better to go back to the initial proposal on that respect, which is to pair `start()` with `complete()`. Thus, for e.g. signature:

```
psa_sign_iop_setup_start()
psa_sign_iop_setup_complete() // repeat while incomplete
psa_sign_iop_setup_update() // repeat as needed; or _hash()
psa_sign_iop_complete() // repeat while incomplete
```

For key agreement:
```
psa_key_agreement_iop_start()
psa_key_agreement_iop_complete() // repeat while incomplete
```

doc/crypto/overview/functionality.rst
```diff
@@ -222,6 +223,169 @@ It is safe to move a multi-part operation object to a different memory location,
...
+See :secref:`hash-mp` for an example of a multi-part operation.
+
+.. _interruptible-operations:
+
+Interruptible operations
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Interruptible operations split a computationally expensive operation into a sequence of function calls, each of which limits the computational progress made. They are useful when responsiveness is critical and the environment does not provide suitable multitasking.
+
+Processing a bounded amount of input in each call can meet this need for some operations. For example, a hash can be computed using a multi-part operation to break the computation into smaller blocks. However, a multi-part operation does not bound computation that is independent of the input size, such as asymmetric signature verification.
```
gilles-peskine-arm: Signature verification might not be a meaningful example because readers might still think about splitting the input. I suggest giving key agreement as the example here.

doc/crypto/overview/functionality.rst
```diff
@@ -222,6 +223,169 @@ It is safe to move a multi-part operation object to a different memory location,
 
 Each type of multi-part operation can have multiple *active* states. Documentation for the specific operation describes the configuration and update functions, and any requirements about their usage and ordering.
 
+See :secref:`hash-mp` for an example of a multi-part operation.
+
+.. _interruptible-operations:
+
+Interruptible operations
```
gilles-peskine-arm, line 217: This section could use a subsection that discusses op counts:

* `xxx_iop_get_num_ops()` functions (still meaningful after `complete()` returns `PSA_SUCCESS`).
* `psa_iop_set_max_ops()` and a brief statement that ops do not have a fixed meaning.

gilles-peskine-arm (changes requested):
I've reviewed the document with the partial implementation in TF-PSA-Crypto in mind. We know that we'll need to update our implementation since it followed a now outdated beta.

Many of my comments apply to multiple operation types. I only noted the first place where I noticed something.

My biggest concern is the verify flow. For verify-message, I don't think providing the signature before the message works out in practice.

athoelke:
> My biggest concern is the verify flow. For verify-message, I don't think providing the signature before the message works out in practice.

Quick response on this:

1. such an observation also affects the already defined verify-message multi-part operation.
2. Deferring the signature is not possible for PureEdDSA, SLH-DSA, LMS/HSS, or XMSS - although it is for ECDSA, RSA-*, HashEdDSA, HashSLH-DSA and all forms of ML-DSA.

A algorithm-agnostic flow requires the signature before the message parts.

doc/crypto/api/keys/management.rst
athoelke: That is certainly a possibility - as we have a multi-step iop object.

doc/crypto/api/ops/signature.rst
athoelke: That limits the algorithm selection: PureEdDSA, SLH-DSA, LMS, and XMSS all require the randomisation element from the signature as part of the hash prefix before the message content.

doc/crypto/figure/interruptible_operation_complex.puml
athoelke: This is an attempt at a 'generic complex interruptible operation'. "Update" is any non-long-running-finalisation action: so this includes both the `psa_sign_iop_update()` and `psa_sign_iop_hash()` fucntions for `psa_sign_iop_t` operations.

doc/crypto/api.db/psa/crypto.h
athoelke: I remember thrashing out numerous variations (although not exhaustively).

On `update`:  We stuck with 'update' for message signature m-p ops and iops due to familiarity with hash.update, cipher.update, mac.update, and aead.update. Not sure there's a good case to make this `setup_update` for this object?

On `complete`: We liked 'complete' as a clear distinction from 'finish' used in other m-p operations, and its alignment with `PSA_ERROR_INCOMPLETE`. But we needed to distinguish the two completion phases for signature iops.

On `setup` vs `start`: We initially used `start`, but then had `start_complete` for the signature iops, which reads very confusingly, and reverted to `setup` and `setup_complete`, relying on the `iop` infix in the function names (`iop_setup` not just `setup`) to provide the reader clue; as m-p operations do not have any m-p infix pattern. If there is a strong feeling for `start` to provide a stronger signal on what is happening, then I am going to argue for `start_setup` and `complete_setup` as a pair because this it eliminate ambiguity from 'setup' as a noun (the operation phase) vs. a verb (the function's action).

So my preference, if we adopt `start`/`complete` pairing, is the same as yours for the simple iops, but the following for the signature iops:

 ```
psa_sign_iop_start_setup()
psa_sign_iop_complete_setup() // repeat while incomplete
psa_sign_iop_update()         // repeat as needed; or _hash()
psa_sign_iop_complete()       // repeat while incomplete
```

File: doc/crypto/api/keys/management.rst (hunk 1 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Fixed

File: doc/crypto/api/keys/management.rst (hunk 4 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Fixed

File: doc/crypto/api/ops/signature.rst (hunk 9 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: Done

File: doc/crypto/api.db/psa/crypto.h (hunk 20 above)
Comment (gilles-peskine-arm) [reply doc/crypto/api.db/psa/crypto.h:920]: > but then had start_complete for the signature iops, which reads very confusingly

Wouldn't that be `setup_start` and `setup_complete`? The sign/verify interface is both multipart and interruptible, and looking more closely, it's both the setup phase and the finish phase of the multipart are both interruptible. The finish phase doesn't need a separate start because all the data is already in the operation. (Although verify would need it if there was a verify-message that takes the signature after the message.)

File: doc/crypto/api/keys/management.rst (hunk 2 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Done

File: doc/crypto/api/keys/management.rst (hunk 5 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Made this an implementation option (by not specifying the exact behaviour in the spec).

File: doc/crypto/api/keys/management.rst (hunk 6 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst:988]: Perhaps - the error is present on both functions exactly for this reason.

File: doc/crypto/api/keys/management.rst (hunk 7 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst:987]: Are you suggesting that this could be a different class of error here?

Comment (athoelke) [reply doc/crypto/api/keys/management.rst:987]: Or is it more of a guidance note for implementations?

File: doc/crypto/api/ops/signature.rst (hunk 10 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: We had previously concluded that if hash() is not called, then update() must be called [at least once], even for an empty message. But since then, the m-p signature operation decided that update() can be called zero times for the empty message signature. So we should probably align with that?

Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: Zero-length messages now permitted via not calling either hash() or update().

File: doc/crypto/overview/functionality.rst (hunk 21 above)
Comment (athoelke) [reply doc/crypto/overview/functionality.rst]: Done

File: doc/crypto/api.db/psa/crypto.h (hunk 20 above)
Comment (athoelke) [reply doc/crypto/api.db/psa/crypto.h:920]: I think we had toyed with `start()` and `start_complete()` as the pair - and never considered usig setup and start at the same time (until now).

File: doc/crypto/api/keys/management.rst (hunk 3 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Now not forbidden.

File: doc/crypto/api/keys/management.rst (hunk 11 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Done

File: doc/crypto/api/ops/signature.rst (hunk 12 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst:3074]: Yes, in theory. The specialized signature algorithms might take a non-hash input: `PSA_ALG_RSA_PKCS1V15_SIGN_RAW` is meant to take a DER-encoded hash. In that case the spec is relaxed to state any byte string the correct length can be used, but we have the same hash-validation error condition on all the other sign/verify hash APIs in case we need a strict rule here.

File: doc/crypto/api/ops/signature.rst (hunk 13 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: Done

File: doc/crypto/api/ops/signature.rst (hunk 14 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst:3141]: INVALID_ARGUMENT would seem to fit best - in cases where a context is being set where the algorithm does not use a context we use this rather than BAD_STATE (which we use for universal API-sequencing errors)?

File: doc/crypto/api/keys/management.rst (hunk 15 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst]: Done

File: doc/crypto/api/ops/signature.rst (hunk 16 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: As per the comment on permitting zero calls for a zero-length message, this morphs into a descriptive comment on what happens if you call psa_sign_iop_complete() before calling hash() or update().

File: doc/crypto/api/ops/signature.rst (hunk 17 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: But with permission for zero input (for zero-length message) this becomes even simpler.

File: doc/crypto/api/ops/key-agreement.rst (hunk 8 above)
Comment (athoelke) [reply doc/crypto/api/ops/key-agreement.rst]: Done

File: doc/crypto/api/ops/signature.rst (hunk 10 above)
Comment (gilles-peskine-arm) [reply doc/crypto/api/ops/signature.rst]: No strong feelings about this one, but a slight preference for allowing not calling either. On the application side, that's easier to handle. On the implementation side, in `complete`, you have to recognize the case where no input function has been called, and it's just as easy to finalize an empty message as to do a bad-state rejection.

File: doc/crypto/api/keys/management.rst (hunk 7 above)
Comment (gilles-peskine-arm) [reply doc/crypto/api/keys/management.rst:987]: Guidance for implementations, and even more for applications. The current wording doesn't make it very clear that reserving is permitted but not mandated.

File: doc/crypto/api/ops/signature.rst (hunk 14 above)
Comment (gilles-peskine-arm) [reply doc/crypto/api/ops/signature.rst:3141]: I'm not sure. On the one hand, as you note, “you can't call this function at this time” is usually `BAD_STATE`. On the other hand, if it's an implementation limitation, then `NOT_SUPPORTED` seems right, and that's usually paired with `INVALID_ARGUMENT` for an algorithm limitation.

File: doc/crypto/api/ops/signature.rst (hunk 10 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: This is now implemented in a new commit. This adds a couple of additional error conditions to `complete()` related to key usage policy and algorithm support for for signing messages.

File: doc/crypto/api/ops/signature.rst (hunk 14 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst:3141]: We don't use BAD_STATE for algorithm-specific limitations elsewhere (KDF, PAKE, signature contexts), and INVALID_ARGUMENT/NOT_SUPPORTED pairing is more typical.

We did not address this in the sign/verify multi-part operations - we only note in the affected algorithms that they are not appropriate for use with the mp operations, but do not specify an error when setting up the operation with that algorithm. Arguably, it could be deferred to the update() for those operations, but fragmenting the message is the entire point of using `psa_sign_operation_t` - whereas using `psa_sign_iop_t` is driven by the need for bounded execution time.

File: doc/crypto/overview/functionality.rst (hunk 22 above)
Comment (athoelke) [reply doc/crypto/overview/functionality.rst:217]: This is currently present in lines 252-256. Is there more that should be said here?

**Comment by athoelke** [comment]
Following discussion - we need to support use cases where the signature is only provided after the message content (required for some streaming protocols), as well as use cases where the signature is provided before the message (which is required for some algorithms).

For `psa_verify_iop_t`, the proposal is to enable a 'late signature' flow, that is only available for some algorithms, and a general 'early signature' flow that works for any algorithm.

Using the current `psa_verify_iop_setup()` function signals the use of the general/early-signature flow, and the rest of the API should be used as currently described.

For the new flow, three additional APIs are provided:

```
psa_status_t psa_verify_iop_setup_deferred_signature(...);
psa_status_t psa_verify_iop_set_signature(...);
#define PSA_ALG_SIGN_SUPPORTS_DEFERRED_SIGNATURE(alg) ...
```

* `psa_verify_iop_setup_deferred_signature()` acts like the current setup function, but without the signature. This indicates that the application will only provide the signature after the message/hash input immediately before completion. If the algorithm or implementation does not support that flow, this is reported as an error response to this function call.
* `psa_verify_iop_set_signature()` supplies the signature to an interruptible verify operation, after all the message/hash input has been provided, and before any call to `psa_verify_iop_complete()`. If called when a signature has already been provided, this returns BAD_STATE.
* If no signature has been provided (either in setup or via `psa_verify_iop_set_signature()`), then `psa_verify_iop_complete()` returns BAD_STATE.
* `PSA_ALG_SIGN_SUPPORTS_DEFERRED_SIGNATURE()` reports whether or not a signature algorithm enables a late-signature flow. This is independent of implememtation-specific support for late flows.

This design, providing a distinct setup function, enables the implementation to detect the application intent explicitly and immediately, and respond with an error if not supported. With a more generic, single setup function without a signature; the application intent can only be inferred when `psa_verify_iop_setup_complete()` is called before providing a signature.

File: doc/crypto/api/keys/management.rst (hunk 7 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst:987]: Do we really want to introduce a single-threaded application model where there is an observable incomplete key object (it does not exist - cannot look the key id up - and it also does exist - cannot create the key id)? What are the costs of requiring atomic behaviour even from these APIs:

* The key identifier can be checked at setup (optional) and an error reported which curtails doing any further redundant work.
* The key identifier is only locked and populated at completion when the key is fully populated.

The obvious costs is that this might result in a late discard of the work done to generate or compute the key if an intervening call has created a key with the same identifier.

Alternatively, we could just make it clear that the correct multi-threaded application guidance also applies to simpler implementations and applications when using interruptible operations:

> The only way to be sure that a key identifier can be allocated to a new key, is to actually create the key. There is always the risk of a race if an application uses a 'test for existence' prior to 'create key' pattern.

File: doc/crypto/api/keys/management.rst (hunk 6 above)
Comment (athoelke) [reply doc/crypto/api/keys/management.rst:988]: done

File: doc/crypto/api/ops/signature.rst (hunk 14 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst:3141]: My mistake - we do specify that the multi-part operations cannot be set up with an algorithm that does not support fragmented messages.

Comment (athoelke) [reply doc/crypto/api/ops/signature.rst:3141]: Now added NOT_SUPPROTED as an option, as per usual for INVALID_ARGUMENT scenarioslike this.

File: doc/crypto/api/ops/signature.rst (hunk 18 above)
Comment (athoelke) [reply doc/crypto/api/ops/signature.rst]: Deferred-signature verification support has been added

File: doc/crypto/api.db/psa/crypto.h (hunk 20 above)
Comment (athoelke) [reply doc/crypto/api.db/psa/crypto.h:920]: I've added another commit that renames the interruptible-initiating calls to have a `start` suffix (Gilles's suggested naming).

* For the simple interruptible operations, this replaces `setup` with `start` - so these no longer have any of the naming patterns associated with multi-part operations.
* For the signature and verification operations, the initiating setup functions have a new `_start` suffix, and the setup completion functions are unchanged. So these operations are both multi-part (have a setup phase) and interruptible (use start/complete).

**Comment by athoelke** [comment]
I think I've managed to address all of the substanive issues raised in the feedback so far. I would appreciate a re-review of the resulting updated PR.

File: doc/crypto/api/keys/management.rst [hunk 23]
```diff
@@ -830,3 +837,546 @@ Key export
...
+    .. retval:: PSA_SUCCESS
+        Success.
+    .. retval:: PSA_ERROR_NOT_SUPPORTED
+        The production parameters are not supported by the implementation.
+    .. retval:: PSA_ERROR_INVALID_ARGUMENT
+        The production parameters are invalid.
+    .. retval:: PSA_ERROR_BAD_STATE
+        The following conditions can result in this error:
+
+        *   The operation state is not valid: it must be active, and `psa_generate_key_iop_complete()` must not have been called.
```
gilles-peskine-arm: `psa_generate_key_iop_custom()` can also fail with `PSA_ERROR_BAD_STATE` if `psa_generate_key_iop_custom()` has already been called.

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,546 @@ Key export
...
+
+    Implementations must reject an attempt to generate a key of size ``0``.
+
+    The following type-specific considerations apply:
+
+    *   For RSA keys (`PSA_KEY_TYPE_RSA_KEY_PAIR`), the public exponent is 65537.
+        The modulus is a product of two probabilistic primes between :math:`2^{n-1}` and :math:`2^n` where :math:`n` is the bit size specified in the attributes.
+
+    After a successful call to `psa_generate_key_iop_start()`, the operation is active.
+    The operation can be configured with custom production parameters by calling `psa_generate_key_iop_custom()`, or completed by calling `psa_generate_key_iop_complete()` repeatedly, until it returns a status code that is not :code:`PSA_OPERATION_INCOMPLETE`.
```
gilles-peskine-arm: It's technically true that `psa_generate_key_iop_custom()` and `psa_generate_key_iop_complete()` are the next two functions that may be used for a normal flow (non-aborted, eventually succeeding). However, the way I understand this sentence is that one may either call `custom()` or the `complete()` loop. It's not clear either here or in the documentation of `custom()` that `custom()` should be followed by a `complete()` loop.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,80 +2785,1062 @@ Multi-part asymmetric signature operations
...
+    .. param:: size_t context_length
+        Size of the ``context`` buffer in bytes.
+
+    .. return:: psa_status_t
+    .. retval:: PSA_SUCCESS
+        Success.
+    .. retval:: PSA_ERROR_BAD_STATE
+        The following conditions can result in this error:
+
+        *   The operation state is not valid: setup must be complete, and no call to `psa_sign_iop_set_context()`, `psa_sign_iop_hash()`, `psa_sign_iop_update()`, or `psa_sign_iop_complete()` has been made.
```
gilles-peskine-arm: There's still an ambiguity here in the second clause, where it's not clear whether the clause describes what caused the error or what must be done to avoid the error. I think it would be simpler to understand if error descriptions always described what causes the error.

```suggestion
        *   The operation state is not valid: setup must be complete, and no call to `psa_sign_iop_set_context()`, `psa_sign_iop_hash()`, `psa_sign_iop_update()`, or `psa_sign_iop_complete()` must have been made.
```

or

```suggestion
        *   The operation state is not valid: the operation is inactive, setup is incomplete, or a call to `psa_sign_iop_set_context()`, `psa_sign_iop_hash()`, `psa_sign_iop_update()`, or `psa_sign_iop_complete()` has been made.
```

(I haven't made a list of where this pattern occurs.)

doc/crypto/api/keys/management.rst
```diff
@@ -830,3 +837,546 @@ Key export
...
+    Once active, the application must eventually terminate the operation. The following events terminate an operation:
+
+    *   A successful call to `psa_generate_key_iop_complete()`.
+    *   A call to `psa_generate_key_iop_abort()`.
+
+    If `psa_generate_key_iop_start()` returns an error, the operation object remains inactive, but its number of *ops* can be reset to zero.
```
gilles-peskine-arm: This still isn't fully accurate: if `psa_xxx_iop_start()` is called on an already active object, it's a `BAD_STATE` error, but the object doesn't “remain inactive”.

doc/crypto/api/keys/management.rst
gilles-peskine-arm: I definitely don't want to implement a key id reservation in TF-PSA-Crypto. I wouldn't mind if reservation was forbidden. But there may be lower-level implementations where persistent key slots are tied more directly to physical addresses where reserving makes more sense, so I'm ok with allowing it in the specification.

doc/crypto/overview/functionality.rst
gilles-peskine-arm: The section introduces the concepts, but I think it should be a bit more concrete. At the very least, it should mention the `get_num_ops` naming convention.

gilles-peskine-arm (changes requested):
Thanks for the updates. I only have a couple of minor nits now, except maybe for one point. I want to think further about verification with and without deferred signature.

doc/crypto/api/keys/management.rst
athoelke: Could we require that binding the application-chosen persistent key id to the key storage slot (whether a software key store or a hardware element slot) is only done at the last step, then non-atomic key creation is not visible in the API. Or can you envisage some implementations where that is untenable?

doc/crypto/api/keys/management.rst
athoelke: I'll review all the sequencing statements affected by adding this function.

doc/crypto/api/ops/signature.rst
athoelke: This could be widespread - I think in all the MP operations, the 'The operation state is not valid:...' conditions for BAD_STATE go on to define what 'valid' means for the function. The statement here mixes requirement tone - 'setup must be complete' - with present tense state tone - 'no call to ... has been made'.

The second alternative is tempting, with the description actually presents the states that are invalid. I would introduce it as 'The operation state is invalid: ...'  - this removes the [existing] ambiguity as to whether the statement that follows is a definition of 'valid states' or of 'not valid states'?

For consistency, this might be a change we should make across all operation functions? - which goes well beyond the remit of this PR. But we could do that for these interruptibles, and follow with a spec-wide edit for the existing multi-parts?

doc/crypto/api/keys/management.rst
athoelke: Quite right, but 'ouch!':

* This affects all the interruptibles in this PR. The statement is only true if the operation was inactive prior to the call.
* All existing multi-part operations suffer from a differently word defect of the same category. They state:

  > If `psa_xxx_setup()` returns an error, the operation object is unchanged.

  Which is also only true if the operation was inactive to start with.
  
 We changed the wording for the interruptibles, because the ops count can be changed by the call. But we have missed the 'If the operation was not inactive, `psa_xxx_setup()` returns `PSA_ERROR_BAD_STATE`, otherwise ...` that should preceed the current statements for all of these APIs.

athoelke: Looks like this warrants an issue to address this omission across all multi-part operations.

doc/crypto/api/keys/management.rst
athoelke: Done

doc/crypto/api/keys/management.rst
athoelke: Done

doc/crypto/overview/functionality.rst
athoelke: Done

athoelke:
I had another thought about the verification and deferred signatures.

For both the multi-part and interruptible verification operations, the operation object wil have to copy some or all of the signature data if it is passed during setup. For PQC signatures, this can be a very sizeable buffer and will be challenging for constrained implementations with no dynamic allocation.

However, although it might work (from a memory management point of view) to have the application deal with the signature memory issue, and pass the signature to both the setup and the finishing phases for an algorithm that requires the signature before the message - is there a security/cryptographic risk in having the application pass what is _meant_ to be the same signature data twice to the implementation?

doc/crypto/api/keys/management.rst
gilles-peskine-arm: I can't rule out an implementation that wants to reserve the slot early, but I can't think of a concrete reason why it would be necessary. So unless you can think of one, I would be in favor of only mentioning the atomic creation behavior in the specification. Technically it wouldn't rule out early slot reservation anyway, since an implementation is allowed to have an implementation-specific error condition where `psa_generate_key_iop_start()` can fail because the target slot is already occupied.

doc/crypto/api/ops/signature.rst
gilles-peskine-arm: I haven't reviewed the `BAD_STATE` descriptions recently. Going by the ones in this PR, the simple ones are ok either way, it's the ones that are half this-causes-an-error and half how-to-avoid-an-error that are confusing. But I would be in favor of aligning everything to this-causes-an-error (in this PR for the new functions, and separately for existing ones), even the ones that aren't confusing on their own.

doc/crypto/api/keys/management.rst
gilles-peskine-arm: Are you going to keep the requirement that a failed setup doesn't change the object?

This requirement was added in 1.1, and [TF-PSA-Crypto hasn't caught up with it](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/855): setup() on an already active multipart operation aborts the operation. We don't mind the requirement from 1.1, but I only noticed very recently and it's low on our priority list.

gilles-peskine-arm:
> For both the multi-part and interruptible verification operations, the operation object wil have to copy some or all of the signature data if it is passed during setup.

I think that for interruptible operations, this memory has to exist anyway between `start()` and the last `complete()` of the finish phase. So it has to be stored in the operation object at some point, directly or indirectly. So for interruptible operations, I think that passing the signature separately will always work.

For multipart operations, it's a different matter: with most algorithms, the signature only needs to exist during the verify-finish phase, which is a single function call.

> is there a security/cryptographic risk in having the application pass what is meant to be the same signature data twice to the implementation?

It's definitely more error prone. I can't think of a direct security risk however, since the inputs of verification are generally not confidential and the output is a single boolean. If the application makes a mistake and passes a different signature, the worst that can happen is that it will get a passing verification without a correct signature. But the application is only harming itself. The surface against glitch attacks is likely much higher, though.

doc/crypto/api/keys/management.rst
athoelke: No, I don't think we ever meant for a call to setup having no effect if the object was active - that should always put the object into an error state, as for any other out-of-sequence call for an active operation.

The current wording was intended to make it clear that a failed setup [on an inactive operation] does not require a call to abort() to reset the operation - but the wording was too broad and we never noticed.

I think these new APIs, and the existing multi-part setup functions need to be reworked to state something along the lines of:

> If `psa_xxx_setup()` returns `PSA_ERROR_BAD_STATE` the operation enters an error state [, and will need to be aborted]. If any other error is returned, the operation object {is unchanged|remains inactive, ...}.

The first sentence mirrors what other operation functions state regarding any error response; the second is a qualified form of the current text for the setup/start functions.

athoelke: Addressed this for the interruptible APIs (see https://github.com/GlobalPlatform/psa-api/pull/375/commits/63f6c68df8585c1816d395373a727801b8102fbd) - borrowing more of the pattern from the existing  multi-part setup functions.

doc/crypto/api/keys/management.rst
athoelke: I've done a bit more than just extend the specific API documentation -

* The pattern for handling persistent key identifiers in interruptible key creation seems to warrant some justification - so I've added a section to the Interruptible operations overview in the Functionality chapter.
* The new key-creation APIs are not atmoic functions. So the term 'key-creation function', previoulsy used throughout the spec is now too narrow. So this is replaced with 'key-creation API', complete with glossary definition.
* The set of functions is now large enough that repeating it in many places is unhelpfully noisy. Most locations just cite the definitive sub-section that describes them all.
* This includes the original list in the Functionality chapter, which is out of step with the overview/summary style of this chapter. The primary list of key-creation APIs is now the section at the top of the Key Management sub-chapter.

Still wondering if we need to fine-tune the wording in the function descriptions a bit more...

doc/crypto/api/ops/signature.rst
athoelke: I've addressed this issue for this PR, and created #381 to rework these clauses for the existing multi-part operations.

doc/crypto/api/keys/management.rst
athoelke: I've also created #380 to address this for the existing multi-part operations.

athoelke:
Another inconsistency has come to light after implementing #381.

The behaviour of the interruptible operation functions when the operation is not in a valid state _for the chosen algorithm_ is to report a `PSA_ERROR_INVALID_ARGUMENT` error (for example trying to update a sign iop with two message fragments, when the algorithm requires that the entire message is provided in a single fragment, or trying to sign a pre-computed hash with an algorithm that can only sign messages).

These are scenarios where the function being called is incorrect for the operation state (including algorithm-specific state) irrespective of the arguments passed to the function. In other multi-part operations, similar scenarios result in a `PSA_ERROR_BAD_STATE` response. For example:

* If an IV is generated or set for a cipher algorithm that does not have an IV
* If a nonce is generated or set before calling `psa_aead_set_lengths()` for an AEAD algorithm that requires the lengths to be explicitly set.
* If a KDF input step is provided out of sequence, or illegally repeated for the chosen algorithm.

I _suspect_ the behaviour (using `PSA_ERROR_INVALID_ARGUMENT`) in the current PR was influenced by the more recent decision on how to handle contexts for signature operations. These do use `PSA_ERROR_INVALID_ARGUMENT`, and implement a default when no context is provided for an algorithm defined to take one - but the justification for this is that not all algorithms define a context, nearly all protocols do not use a context, and there is an obvious default that is defined to work as required in algorithms that take them.

There are some scenarios in the new interruptible operations (and possibly in the multi-part sign/verify operations?) which might be better served by using `PSA_ERROR_BAD_STATE` to report that the API being called is not appropriate for the current operation?

athoelke:
Another unusual aspect to the sign and verify iop API is that the permission check for the keys is deferred until the application calls the `hash()`, `update()`, or `complete()` functions. This is because we have distinct usage policies for signing (or verifying) hashes and messages. Every other operation checks the key policy at the time the key is provided to the operation.

This is not ideal, both from an application error discovery point of view, or from an implementation point of view (which often combines key lookup with policy verification).

I propose to check for the appropriate SIGN_MESSAGE/VERIFY_MESSAGE usage when the operation is set up - noting that SIGN_HASH usage always implies SIGN_MESSAGE permission as well. Then there is a second permission check if the applicaiton uses the `hash()` function to sign or verify a pre-computed hash, to ensure that the key had the stronger SIGN_HASH/VERIFY_HASH permission. As a result, the permission checks on `update()`, `set_signature()` or `complete()` are no longer needed.

Although this still defers part of the permission check in some uses, it provides a better compromise without introducing separate setup APIs for hash and message interruptible signature operations.

athoelke:
I have added a few more commits to regularise the following:

* Signatures that are invalid are now only reported when completing the interruptible verification, using `PSA_ERROR_INVALID_SIGNATURE` - as is done in other signature verification functions and operations.
* Interruptible operation function calls that are not valid for the operation state (including the selected algorithm) are now `PSA_ERROR_BAD_STATE` errors.
* Key usage policy verification for signature and verification is clarified: the only deferred usage check is when using a pre-computed hash in the new operations. This approach is based on the behaviour of the message and hash signing/verifying usage policies: e.g. a key with `PSA_KEY_USAGE_SIGN_HASH` automatically also has `PSA_KEY_USAGE_SIGN_MESSAGE`.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,80 +2785,1070 @@ Multi-part asymmetric signature operations
...
+
+    The application must complete the setup of the operation before calling this function.
+
+    For message-signature algorithms that process the message data multiple times when verifying a signature, `psa_verify_iop_update()` must be called exactly once with the entire message content. For signature algorithms that only process the message data once, the message content can be passed in a series of calls to `psa_verify_iop_update()`.
+
+    After input of the message, the verification operation can be completed by calling `psa_verify_iop_complete()` until it returns a status code that is not :code:`PSA_OPERATION_INCOMPLETE`.
+
+    If this function returns an error status, the operation enters an error state and must be aborted by calling `psa_verify_iop_abort()`.
+
+.. function:: psa_verify_iop_set_signature
```
gilles-peskine-arm, line 3595: I'm not sure I agree with c91bbec5b084fa5b4467896735d78f3ff1857266. Why can't the function that receives the signature signal an error if it can already tell that the signature is invalid? For example, suppose `psa_verify_iop_setup_start()` or `psa_verify_iop_set_signature()` is called with a signature that is larger than what it can copy into the operation object. The only thing it can now do is to set an “invalid signature” flag in the operation object, which needs to stay until the call to `psa_verify_iop_complete()`, possibly past `psa_verify_iop_update()` calls. Or suppose that `psa_verify_iop_setup_start()` is called with an algorithm where the signature contains a nonce which needs to be hashed before the message, but the signature is too short and the implementation can't extract a nonce: what is `psa_verify_iop_update()` supposed to do?

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,80 +2785,1070 @@ Multi-part asymmetric signature operations
...
+
+    .. return:: psa_status_t
+    .. retval:: PSA_SUCCESS
+        Success.
+    .. retval:: PSA_ERROR_BAD_STATE
+        The following conditions can result in this error:
+
+        *   The operation is not active.
+        *   The operation setup has not completed successfully.
+        *   `psa_verify_iop_hash()`, `psa_verify_iop_set_signature()`, or `psa_verify_iop_complete()` has already been called.
```
gilles-peskine-arm: `psa_verify_iop_setup_start()` supports algorithms where passing the signature can be deferred. This means that the operation object must have room to store the signature while the operation is ongoing. Then I don't see why `psa_verify_iop_set_signature()` cannot be called before supplying the message.

We may want to allow the implementation to overlap the signature storage with the message hash state in the operation object, and perhaps even to store only part of the signature (if the algorithm allows a first part of the verification to be done quickly, and a second part to be done more slowly but with less long-term memory). But if so, the non-deferred `psa_verify_iop_setup_start()` must reject algorithms that support deferred signature.

doc/crypto/api/ops/signature.rst
```diff
@@ -2750,80 +2785,1070 @@ Multi-part asymmetric signature operations
...
+        Number of *ops* that the operation has taken so far.
+
+    After the interruptible operation has completed, the returned value is the number of *ops* spent on the entire operation. The value is reset to zero by a successful call to `psa_verify_iop_setup_start()`, `psa_verify_iop_setup_deferred_signature_start()`, or `psa_verify_iop_abort()`.
+    A failed call to either setup function can also reset the value to zero.
+
+    This function can be used to tune the value passed to `psa_iop_set_max_ops()`.
+
+    The value is undefined if the operation object has not been initialized.
+
+.. function:: psa_verify_iop_setup_start
```
gilles-peskine-arm, line 3336: Alternative proposal for verification:

* `psa_verify_iop_setup_start()` does not take a signature argument.
* There is no separate function `psa_verify_iop_setup_deferred_signature_start()`.
* If the algorithm supports deferred signature verification, you may call `psa_verify_iop_set_signature()` at any time after `setup_start()` and before `complete()`.
* If the algorithm does not support deferred signature, you must call `psa_verify_iop_set_signature()` before `psa_verify_iop_setup_complete()`.

This feels to me like it's a little easier to both use and implement.

It means the operation object must have room for the signature, but [that's already the case](https://github.com/GlobalPlatform/psa-api/pull/375#discussion_r3854991287).

gilles-peskine-arm (changes requested):
Mostly LGTM, but there are still a few kinks in the verify flow.

doc/crypto/api/ops/signature.rst
athoelke: This change was done to align with the behaviour of multi-part asymmetric verification (in v1.5 - see [`psa_verify_setup()`](https://arm-software.github.io/psa-api/crypto/1.5/api/ops/signature.html#c.psa_verify_setup)).

I think we went in that direction to simplify application code so that there is a single location where a signature mismtach is erported (and possibly to reduce the application-level differences in control flow when corrupted signatures are presented?).

However, I can understand that it requires additional implementation code/state to handle what are clearly invalid signatures, whilst appearing to permit the continuation of the operation - particularly when the signature is defined to include a nonce for the message processing...

doc/crypto/api/ops/signature.rst
athoelke: The intention here is good:

* The early signature API is algorithm-agile
* The late signature API is necessary for streaming use cases

So this is good for applications. But the implementation cost for the former can be substantial, as there is no room for space optimization. A couple of options:

1. Make the API split stronger - require that algorithms that support deferred signatures can only be used with the late-signature APIs.
2. Make the implementation support decision more evident - explicitly list something like 'implementation does not support use of early-signature API with this algorithm' under `PSA_ERROR_NOT_SUPPORTED`.

The former makes it explicit that for algorithm agility, an application has to code for both flows depending on the algorithm and the SUPPORT macro response. The latter _suggests_ that algorithm-agile applications might need to do that for full portability across implementations.

doc/crypto/api/ops/signature.rst
athoelke: This was an option we did discuss - and it does look neater as an API design.

This approach means that:

1. The signature is never required to complete the setup for any verification algorithm.
2. We are happy with reporting `PSA_ERROR_BAD_STATE` from `psa_verify_iop_update()` - because `psa_verify_iop_set_signature()` has not been called [as required for the algorithm] - being the first indiciation to an application that their algorithm is not compatible with the call flow. (Obviously, the app should have used the support macro to check for this...)

athoelke: If we went down this route instead, then I'd be inclined to change the position w.r.t. whether an early signature is always acceptable. I would explicitly document that an agile application must use the macro to select between supplying the signature early vs supplying the signature late.

doc/crypto/api/ops/signature.rst
athoelke: *Decision*: INVALID_SIGNATURE is permitted at the point the signature is provided when it is structurally invalid and cannot be used to perform the verification.

doc/crypto/api/ops/signature.rst
athoelke: Ooops I misread the proposal. (1) is not true - for early signatures @gilles-peskine-arm suggested that the signature must be provided before setup_complete().

For the IOP, a separate set_signature() function looks tidier as a new API, and due to the interruptible behaviour the implementation will have to store significant parts of the signature for some of the lifetime of the operation. In comparison, the multi-part message verification API (#379) benefits significantly from the deferred signature being passed in the finish() function.

*Decision*: Change the API design to have a single setup_start() function without signature, and a set_signature() function that can be called late for appropriate algorithms.

doc/crypto/api/ops/signature.rst
athoelke: *Decision*: For the IOP, the implementation will have to store some/all of the signature for some of the interruptible operation lifetime. Thus requiring the implementation support the signature being provided before the message for all supported algorithms is acceptable for the implementations, and better for application design.

The approach for the multi-part message verify operation will be different, as the non-interruptible nature changes the implementation implications.

gilles-peskine-arm (approved):
LGTM for the overall design.

I've done a detailed review of most of the content, but I haven't verified that the last commit 84ad096c9c2ec615b23f089249810a6e9a4db5ec is complete.

doc/crypto/api/ops/signature.rst
athoelke: Now fixed, and also in the multi-part verification operation.

doc/crypto/api/ops/signature.rst
athoelke: Implemented in this PR

athoelke:
Rebased to resolve conflict in history.rst with upstream main branch.

PR #382, opened 2026-09-08
Extend the list of invalid hash algorithms for RSA PKCS#1v1.5 including Blake2

valeriosetti:
Blake2 hash algorithms do not have a standardized OID, so they cannot be used for RSA PKCS#1v1.5. The documentation of these hash algorithms already states that, but the documentation of PSA_ALG_RSA_PKCS1V15_SIGN() misses this info. This commit resolves this gap.


athoelke (approved):
Thank you. We missed adding the BLAKE2 algorithms to this list when introducing these hashes.

LGTM

